PR #18, opened 2021-06-16
Modify boot node endpoint

imstar15:
testnet.oak.tech -> rpc.testnet.oak.tech


chrisli30 (approved):
okay yeah rpc endpoint

PR #32, opened 2021-11-11
Fix tests

imstar15:
All the tests passed.

"SKIP WASM BUILD=1 cargo test should be changed to "cargo test".
Otherwise the test of "test_create_development_chain_spec", "test_create_local_testnet_chain_spec" will not pass.


irsal (approved):
Nice!

PR #33, opened 2021-11-22
Add README.md in session_keys folder

imstar15:


chrisli30 (changes requested):
@imstar15 where do those additional changes come from? Can you be careful about the PR?

CleanShot 2021-11-22 at 08 59 15

chrisli30 (approved):
Okay, my mistake. All commits are for this PR. The keystore files are removed so I'm approving it. 

@irsal let's verify Charle's new script for the key configurations.

PR #37, opened 2021-12-08
Add pallet-treasury

imstar15:


chrisli30 (approved):
Code look good to me. 👍 

Can we test it locally and add screenshot of the treasury here before merging it?

irsal:
A couple of questions:

- How do we seed the treasury with funds?
- Does the council manage the treasury?
- Who should be part of the council?

imstar15:
> A couple of questions:
> 
> * How do we seed the treasury with funds?

I haven't tried how to seed funds into treasure. Maybe we can initialize the configuration in the chain spec, or transefer balance to the treasury account.

In polkadot, the treasury is funded from different sources: slashing, transaction fees, staking inefficiency, parathreads.

https://wiki.polkadot.network/docs/learn-treasury#funding-the-treasury

> * Does the council manage the treasury?
Yes. In the early stage of network startup, the Sudo account can also temporarily manage treasure.

> * Who should be part of the council?
Council members are selected through a referendum on the Democracy module. 

> * How do we seed the appropriate accounts from the beginning of the chain for the members of the council?
https://github.com/paritytech/polkadot/blob/37fc426f7090e2e1e132645a955f880751ba8a69/node/service/src/chain_spec.rs#L348

> * What is the difference between this council and the technical council?
Council: 
* Controlling the treasury.
* Proposing sensible referenda.
* Cancelling uncontroversially dangerous or malicious referenda.
* Electing the technical committee.
More details:  https://wiki.polkadot.network/docs/learn-governance#council

Technical committee:
The Technical Committee can, along with the Council, produce emergency referenda, which are fast-tracked for voting and implementation. These are used for emergency bug fixes or rapid implementation of new but battle-tested features into the runtime.
https://wiki.polkadot.network/docs/learn-governance#technical-committee

runtime/neumann/src/lib.rs
```diff
@@ -583,7 +586,81 @@ impl pallet_collator_selection::Config for Runtime {
 	type WeightInfo = ();
 }
 
-/// Configure the pallet template in pallets/template.
+parameter_types! {
+	pub const CouncilMotionDuration: BlockNumber = 5 * DAYS;
```
chrisli30, line 590: The parameters are good for Turing Network, but long for development and Neumann. We don’t have separate chain_specs for different networks yet, do we?

chrisli30 (approved):
Code looks good to me. If the automated tests can pass, let’s merge the code.

runtime/neumann/src/lib.rs
```diff
@@ -583,7 +588,86 @@ impl pallet_collator_selection::Config for Runtime {
...
+		EnsureRoot<AccountId>,
+		pallet_collective::EnsureProportionAtLeast<_3, _5, AccountId, CouncilCollective>,
+	>;
+	type RejectOrigin = EnsureOneOf<
+		AccountId,
+		EnsureRoot<AccountId>,
+		pallet_collective::EnsureProportionMoreThan<_1, _2, AccountId, CouncilCollective>,
+	>;
+	type Event = Event;
+	type OnSlash = ();
```
rhuttman, line 659: why `()` and not `Treasury`?

runtime/neumann/src/lib.rs
imstar15: Yes, we need to separate different runtime configurations for different networks.

runtime/neumann/src/lib.rs
imstar15: proposal bond: Of the beneficiary amount, at least 5.00% would need to be put up as collateral.
image

**type OnSlash = ();**
It means just deducting the amount of the proposer bond.

**type OnSlash = Treasury;**
It means deducting the amount of the proposer bond and sending it to the treasure. .

examples:
In Acala and Karura's runtime code, use `type OnSlash = Treasury;`
https://github.com/AcalaNetwork/Acala/pull/1259/files#diff-c4e09516778f20bc59aa942a45c3632683b1202784cc82fdd368101f705734e9L578

In Mandala's runtime code, use type OnSlash = ();https://github.com/AcalaNetwork/Acala/blob/8beb6f8659ca90def7e5c3949bf5414d8bc8befe/runtime/mandala/src/lib.rs#L578

imstar15: I think Neumann is a testnet similar to Mandala, so you can refer to Mandala's code.

chrisli30: This different between () and Treasury is trivial to users. Since we don’t have separate code base for each network yet, let’s all use **type OnSlash = Treasury** for simplicity here.

chrisli30 (approved):

PR #38, opened 2021-12-09
Delete session keys.

imstar15:


irsal:
@imstar15 - what's scripts/session_keys/.gitkeep ?

imstar15:
> @imstar15 - what's scripts/session_keys/.gitkeep ?

A GITKEEP file is an empty file that Git users create so that a Git repository preserves an otherwise empty project directory. By convention, empty directories that contain no files are not committed to Git repositories. However, a single GITKEEP file in a directory causes Git to preserve the directory in the repository.

chrisli30 (changes requested):
Yes, let’s remove the .gitkeep, and this folder completely.

chrisli30 (approved):

PR #39, opened 2021-12-09
Add pallet-sudo

imstar15:


node/src/chain_spec.rs
```diff
@@ -90,6 +90,8 @@ pub fn development_config() -> ChainSpec {
 						get_collator_keys_from_seed("Bob"),
 					),
 				],
+				// Sudo account
+				get_account_id_from_seed::<sr25519::Public>("Alice"),
```
imstar15, line 94: Here is a temporary Sudo account, which needs to be modified later.

node/src/chain_spec.rs
```diff
@@ -144,6 +146,8 @@ pub fn local_testnet_config() -> ChainSpec {
...
```
imstar15, line 150: Here is a temporary Sudo account, which needs to be modified later.

chrisli30: Why do we need to write a seed phrase in the chain config? It doesn’t make sense.

I just need a wallet addres to be a sudo. I shouldn’t need to give you my seed phrase.

chrisli30 (changes requested):

chrisli30 (changes requested):

rhuttman (approved):
lgtm

chrisli30 (approved):
My questions were addressed so looking good 👍

PR #42, opened 2021-12-22
Test coverage

imstar15:
When the PR is submitted, the coverage is automatically checked.


Cargo.toml
```diff
@@ -10,64 +10,3 @@ members = [
...
-# Note that this does **not** affect crates that depend on Substrate. In other words, if you add
-# a dependency on Substrate, you have to copy-paste this list in your own `Cargo.toml` (assuming
-# that you want the same list). This list is only relevant when running `cargo build` from within
-# the Substrate workspace.
-#
-# If you see an error mentioning "profile package spec ... did not match any packages", it
-# probably concerns this list.
-#
-# This list is ordered alphabetically.
-[profile.dev.package]
```
imstar15, line 29: Remove the compile optimization option in dev mode, it will make the grcov command fail.
The error is:
```
Unsupported instrumentation profile format version
```

Refer to the substrate-parachain-template project, which does not have these configuration items in its Cargo.toml.
https://github.com/substrate-developer-hub/substrate-parachain-template/blob/master/Cargo.toml

.github/workflows/check.yml
```diff
@@ -59,3 +59,36 @@ jobs:
...
+          rustup default stable
+          rustup update nightly
+          rustup update stable
+          rustup target add wasm32-unknown-unknown --toolchain nightly
+          rustup component add llvm-tools-preview
+          cargo install grcov
+
+      - name: Test coverage
+        run: |
+          export CARGO_INCREMENTAL=0
```
imstar15, line 84: The following commands are referenced:
https://github.com/paritytech/polkadot/blob/83e5955b83870c0b107af37934d9f8a4ecf2c63d/doc/testing.md#coverage

irsal (approved):
👍

.github/workflows/check.yml
```diff
@@ -59,3 +59,36 @@ jobs:
...
+          export SKIP_BUILD_WASM=true
+          export BUILD_DUMMY_WASM_BINARY=true
+          export LLVM_PROFILE_FILE="llvmcoveragedata-%p-%m.profraw"
+          export WASM_TARGET_DIRECTORY=/tmp/wasm
+          cargo +nightly build
+          export RUSTFLAGS="-Zinstrument-coverage"
+          rm -r target/debug
+          cargo +nightly test --all
```
imstar15, line 92: The test coverage workflow contains the commands for automatic testing, so the run-tests workflow can be removed. What do you think?

@irsal

irsal: For the test coverage workflow, if any of the tests fail, does the job fail?

imstar15: Yes, it will fail.

irsal: Ok, then we can remove `run-tests` workflow

PR #43, opened 2021-12-23
Fix the problem that mac m1 cannot run docker image.

imstar15:
Refer to:
https://github.com/paritytech/polkadot/blob/master/scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile
https://github.com/paritytech/polkadot/blob/master/scripts/dockerfiles/polkadot/build.sh

Usage:
```
sh docker/neumann/build.sh
```

**!!!This script will read the version configured in node/Cargo.toml as the version of the docker image. Therefore, when updating the node's code, remember to update the version field of this file.**

Building the image is possible on any host having docker installed and will build OAK inside Docker. That also means that no Rust toolchain is required on the host machine for the build to succeed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,6 @@
+.github
+.git
+.vscode
 target
 data
 !resources
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
-*
+target
+data
 !resources
 !target/release/oak
 !target/release/neumann-collator
-
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4405,7 +4405,7 @@ dependencies = [
 
 [[package]]
 name = "oak-node"
-version = "0.1.0"
+version = "1.2.1"
 dependencies = [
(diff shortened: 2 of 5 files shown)


imstar15:
Please do not build the image in the Mac M1 environment, because you will encounter problems.

```
#11 172.4    Compiling proc-macro2 v1.0.32
#11 172.4    Compiling unicode-xid v0.2.2
#11 172.4    Compiling syn v1.0.82
#11 172.4    Compiling libc v0.2.108
#11 184.8 Killed
------
executor failed running [/bin/sh -c cargo build --locked --release]: exit code: 137
```

chrisli30 (approved):
According to Charles, the reason that Mac M1 chip couldn’t run our Docker image is that we built the binary file on Linux and copied over to the Docker image. That was wrong.

This change is to move the binary file building process into the Docker container, out of which the binary outcome could run on Mac M1 chip.

irsal (approved):
Makes sense

PR #50, opened 2022-01-10
Upgrade dependences to polkadot-v0.9.13

imstar15:
FYI: 
1. The `--parachain-id` parameter in the CLI command has been removed. 
**!!!Please configure `parachain_id` through the exported `chain spec` file.**

2. I added the `rust-toolchain.toml` file to specify the version and target of the rust toolchain.
If the corresponding component is not installed on the machine, it will be installed automatically at build time.

Refer to:
https://github.com/paritytech/cumulus/pull/739


imstar15:
The latest rust nightly has a bug that will cause build polkadot-v0.9.12 to fail, so the check action will fail.
https://githubmate.com/index.php/repo/bytecodealliance/rustix/issues/142

It is recommended to use version nightly-2021-12-15.

node/src/chain_spec.rs
```diff
@@ -12,6 +12,7 @@ const SS_58_FORMAT: u32 = 42;
 const TOTAL_TOKENS: u128 = DOLLAR * 1_000_000_000;
 static RELAY_CHAIN: &str = "rococo-local";
 static NEUMANN_RELAY_CHAIN: &str = "rococo-testnet";
+const DEFAULT_PARA_ID: u32 = 1000;
```
irsal, line 15: 2000

chrisli30:
> The latest rust nightly version on Ubuntu has a bug that will cause build polkadot-v0.9.13 to fail, so the check action will fail. https://githubmate.com/index.php/repo/bytecodealliance/rustix/issues/142
> 
> It is recommended to use version nightly-2021-12-15.

I think we can lock the version "nightly-2021-12-15" for the CI to use. There‘s no need to upgrade Rust library real-time.

irsal:
I'll take over this PR as well.

imstar15:
I added the rust-toolchain.toml file to specify the version and target of the rust toolchain.
If the corresponding component is not installed on the machine, it will be installed automatically at build time.

pallets/automation-time/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "pallet-automation-time"
 description = "Pallet for scheduling and running tasks in the future."
 version = "0.0.0"
-edition = "2021"
+edition = "2018"
```
rhuttman, line 5: 2018?

rhuttman (approved):

pallets/automation-time/Cargo.toml
irsal: We haven't effectively switched to rust 2021. Depending on your toolchain, it might force you to update the rust config to `edition2021`. We should move to 2021 in a future pr (like soon)

PR #51, opened 2022-01-10
Schedule a task to send an event with custom text

rhuttman:


pallets/automation-time/src/lib.rs
```diff
@@ -15,26 +15,186 @@
...
+		/// * `time`: The unix standard time in seconds for when the task should run.
+		/// * `message`: The message you want the event to have.
+		///
+		/// # Errors
+		/// * `InvalidTime`: Time must end in a whole minute.
+		/// * `PastTime`: Time must be in the future.
+		/// * `EmptyMessage`: The message cannot be empty.
+		/// * `DuplicateTask`: There can be no duplicate tasks.
+		/// * `TimeSlotFull`: Time slot is full. No more tasks can be scheduled for this time.
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(2) + T::DbWeight::get().reads(2))]
```
irsal, line 132: how'd we calculate this weight?

rhuttman: really its a placeholder. I will get the real values in phase 4. Notice that this pallet isn't even exposed to the runtime

pallets/automation-time/src/lib.rs
```diff
@@ -15,26 +15,186 @@
...
+			Ok(().into())
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		/// Get the relevant time slot.
+		///
+		/// In order to do this we get the most recent timestamp from the chain. Then convert
+		/// the ms unix timestamp to seconds. Lastly, we bring the timestamp down to the last whole minute.
+		fn get_time_slot() -> u64 {
```
irsal, line 176: `get_current_time_slot`

pallets/automation-time/src/lib.rs
```diff
@@ -15,26 +15,186 @@
...
+		/// * `EmptyMessage`: The message cannot be empty.
+		/// * `DuplicateTask`: There can be no duplicate tasks.
+		/// * `TimeSlotFull`: Time slot is full. No more tasks can be scheduled for this time.
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(2) + T::DbWeight::get().reads(2))]
+		pub fn schedule_notify_task(
+			origin: OriginFor<T>,
+			time: u64,
+			message: Vec<u8>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
```
irsal, line 138: why does this need to be signed? Also, is there a expiry for `ensure_signed`? ie can we still use this function in 2 years?

irsal (approved):
nits & questions; got it to run in my machine, no extrinsics exposed yet (which is to be expected)

pallets/automation-time/src/lib.rs
rhuttman: We don't want someone to schedule a task for someone else. All this does is throw an error if it't signed else return  the accountId of the caller. 

If you are thinking about how we will tell other chains when someone signed for something on our chain months earlier we will not be able to use this concept sadly.

PR #57, opened 2022-01-16
Scheduled tasks overflow support

rhuttman:


pallets/automation-time/src/tests.rs
```diff
@@ -253,19 +250,167 @@ fn run_tasks_completes_some_tasks() {
...
+				Event::AutomationTime(crate::Event::Notify { message: message_two.clone() }),
+			]
+		);
+		assert_eq!(expected_task_id, AutomationTime::get_overflow_tasks().unwrap()[0]);
+		assert_eq!(1, AutomationTime::get_overflow_tasks().unwrap().len());
+	})
+}
+
+#[test]
+fn trigger_tasks_pushes_some_to_overflow() {
```
irsal, line 365: probably unlikely, what about 2 overflow consecutive? i know we execute the overflow tasks first, but what happens if we have too much volume when there's constant overflow?

irsal: overflows can also be weight based right?

irsal (approved):

pallets/automation-time/src/tests.rs
rhuttman: the only way to protect against that is by adding buffers to our max scheduled tasks. An example would be allowing only 80% of the max tasks to be scheduled for a given slot. Since we cannot know a backup will occur ahead of time, we need to be very careful when we define our limits and safety buffers

pallets/automation-time/src/lib.rs
```diff
@@ -264,15 +272,21 @@ pub mod pallet {
...
 
 	impl<T: Config> Pallet<T> {
-		/// Get the relevant time slot.
+		/// Based on the block time return the time slot and if it's the last block in the slot.
 		///
-		/// In order to do this we get the most recent timestamp from the chain. Then convert
-		/// the ms unix timestamp to seconds. Lastly, we bring the timestamp down to the last whole minute.
-		fn get_current_time_slot() -> u64 {
+		/// In order to do this we get the most recent timestamp from the block. Then convert
+		/// the ms unix timestamp to seconds. , we bring the timestamp down to the last whole minute.
```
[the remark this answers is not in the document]
rhuttman: I will fix this when i go through all the comments

PR #59, opened 2022-01-19
Adding event for missing task

rhuttman:


irsal (approved):

pallets/automation-time/src/tests.rs
```diff
@@ -209,16 +229,35 @@ fn cancel_must_be_owner() {
...
-		assert_ok!(AutomationTime::cancel_task(Origin::signed(ALICE), task_id,));
 		assert_noop!(
 			AutomationTime::cancel_task(Origin::signed(ALICE), task_id),
 			Error::<Test>::TaskDoesNotExist,
 		);
 	})
 }
 
+#[test]
+fn cancel_task_not_found() {
```
irsal, line 243: I don't get what this does.

rhuttman: there is a weird edge case were somehow the task exists but isn't scheduled or in the overflow. There is no way this state should ever happen but if it does I want to know about it

irsal (approved):
quick que

PR #70, opened 2022-02-08
adjust action notify enum

justinzhou93:


pallets/automation-time/src/lib.rs
```diff
@@ -71,11 +71,17 @@ pub mod pallet {
...
+	pub struct NotificationAction {
+		message: Vec<u8>,
+	}
+
 	/// The enum that stores all action specific data.
 	#[derive(Debug, Eq, PartialEq, Encode, Decode, TypeInfo)]
 	#[scale_info(skip_type_params(T))]
 	pub enum Action {
-		Notify(Vec<u8>),
+		Notify(NotificationAction),
```
rhuttman, line 84: hmm I was just thinking of doing this in-line, but i guess a struct is fine

rhuttman (commented):
I would hope at least 1 test failed due to this

pallets/automation-time/src/lib.rs
justinzhou93: curious how i even do this inline. looked into it a bit and it looks like structs cannot be anonymous

rhuttman: same thing as the events right below it

rhuttman (approved):

justinzhou93:
code coverage is wrong, since it does it based on lines, not on what code has been run

PR #73, opened 2022-02-09
OnF neumann testnet parachain creation script

justinzhou93:


resources/neumann-testnet-parachain.yaml
```diff
@@ -0,0 +1,67 @@
...
+            - key: --chain
+              file: "rococo-testnet.json"
+            - key: --bootnodes # Relay chain p2p address
+              value: /dns4/node-6896992096614604800-0.p2p.onfinality.io/tcp/23160/ws/p2p/12D3KooWM1vTue78y2ZjxTjZKgYEPR8gSNFtTda3KETgp3WAhCxK
+        collator:
+          args:
+            - key: "--force-authoring"
+            - key: --chain
+              file: "neumann.json"
+            - key: --parachain-id
```
rhuttman, line 28: i think youi meant to remove this

resources/neumann-testnet-parachain.yaml
```diff
@@ -0,0 +1,67 @@
...
+              value: wasm
+            - key: --chain
+              file: "rococo-testnet.json"
+            - key: --bootnodes # Relay chain p2p address
+              value: /dns4/node-6896992096614604800-0.p2p.onfinality.io/tcp/23160/ws/p2p/12D3KooWM1vTue78y2ZjxTjZKgYEPR8gSNFtTda3KETgp3WAhCxK
+validator:
+  count: 2
+  node:
+    nodeName: neumann-parachain-testnet-collator-jzia5
+    nodeType: collator
```
rhuttman, line 44: does this work bnow?

resources/neumann-testnet-parachain.yaml
justinzhou93: ahh yes yes, thank you. thanks for the catch haha. god knows we don't want to ask OnF this thing again

resources/neumann-testnet-parachain.yaml
justinzhou93: it doesn't by itself cause of hte bootnode. you need to manually create the bootnode and connect the P2P address and remove the validator param. after that, it works. but it saves like 10 min each time from creating the network and collators manually haha

rhuttman: can we add a comment somewhere about hpw this is just to save time but won't work without UI edits

rhuttman (approved):

resources/neumann-testnet-parachain.yaml
justinzhou93: for sure - lemme add! also will def add in the doc to run the neumann testnet

resources/neumann-testnet-parachain.yaml
```diff
@@ -0,0 +1,68 @@
...
+              file: "rococo-testnet.json"
+            - key: --bootnodes # Relay chain p2p address
+              value: /dns4/node-6896992096614604800-0.p2p.onfinality.io/tcp/23160/ws/p2p/12D3KooWM1vTue78y2ZjxTjZKgYEPR8gSNFtTda3KETgp3WAhCxK
+        collator:
+          args:
+            - key: "--force-authoring"
+            - key: --chain
+              file: "neumann.json"
+            - key: --collator
+            - key: --bootnodes # Relay chain p2p address
```
irsal, line 32: # Parachain p2p address

irsal (approved):

PR #74, opened 2022-02-10
adding transfer scheduler

justinzhou93:


pallets/automation-time/src/lib.rs
```diff
@@ -121,6 +140,18 @@ pub mod pallet {
 		/// The time each block takes.
 		#[pallet::constant]
 		type SecondsPerBlock: Get<u64>;
+
+		/// The data type for balance amounts.
```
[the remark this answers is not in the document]
rhuttman: you should be passing in the type: https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L203
this line makes it a constant that is expecting a number to be passed in

pallets/automation-time/src/lib.rs
```diff
@@ -121,6 +140,18 @@ pub mod pallet {
...
+		#[pallet::constant]
+		type Balance: Get<u128>;
+
+		/// Access to Balance Pallet.
+		#[pallet::constant]
```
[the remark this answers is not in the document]
rhuttman: it makes it expect a value: https://docs.substrate.io/v3/runtime/macros/

pallets/automation-time/src/lib.rs
```diff
@@ -245,6 +285,54 @@ pub mod pallet {
...
+			provided_id: Vec<u8>,
+			time: UnixTime,
+			recipient_id: T:: AccountId,
+			amount: u128, // TODO: jzhou fix to be type runtime::Balance
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			let recipientExists = account_exists(recipient_id);
+			
+			// check for greater than existential deposit
+			ensure!(amount > EXISTENTIAL_DEPOSIT, Err(<Error<T>>::InvalidAmount))?;
```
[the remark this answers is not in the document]
rhuttman: Ensure returns a result. The first param is the condition, the second is the error that is returned in the condition fails. Therefore you want it at the end of the ensure.

The ? says to pass the error up IF the fcn call it's after returns a result AND the fcn it is in returns a result

pallets/automation-time/src/lib.rs
```diff
@@ -526,5 +644,20 @@ pub mod pallet {
 			}
 			Ok(task_id)
 		}
+
+		pub fn validate_and_schedule_task(
```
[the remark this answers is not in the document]
rhuttman: you could have this take in an action and save the task at the end. plus fire the event

pallets/automation-time/src/weights.rs
```diff
@@ -62,6 +63,15 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(3 as Weight))
 			.saturating_add(T::DbWeight::get().writes(2 as Weight))
 	}
+	// Storage: Timestamp Now (r:1 w:0)
```
rhuttman, line 66: where did this come from?

runtime/neumann/src/lib.rs
```diff
@@ -682,6 +682,7 @@ parameter_types! {
 	pub const MaxBlockWeight: Weight = MAXIMUM_BLOCK_WEIGHT;
 	pub const MaxWeightPercentage: Perbill = SCHEDULED_TASKS_INITIALIZE_RATIO;
 	pub const SecondsPerBlock: u64 = MILLISECS_PER_BLOCK / 1000;
+	pub const ExistentialDeposit: Balance = EXISTENTIAL_DEPOSIT;
```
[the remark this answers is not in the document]
rhuttman: ummmm idk this one

runtime/neumann/src/lib.rs
```diff
@@ -692,6 +693,9 @@ impl pallet_automation_time::Config for Runtime {
 	type MaxWeightPercentage = MaxWeightPercentage;
 	type SecondsPerBlock = SecondsPerBlock;
 	type WeightInfo = pallet_automation_time::weights::AutomationWeight<Runtime>;
+	type Balance = Balance;
+	type Currency = Balances;
```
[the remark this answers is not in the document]
rhuttman: you need to import the currency trait in the automation pallet

runtime/neumann/src/lib.rs
```diff
@@ -692,6 +693,9 @@ impl pallet_automation_time::Config for Runtime {
...
+	type ExistentialDeposit = ExistentialDeposit;
```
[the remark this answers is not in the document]
rhuttman: the #[pallet::constant] wraps the value. read the macro for it: https://docs.substrate.io/v3/runtime/macros/

pallets/automation-time/src/weights.rs
justinzhou93: when i didn't have `#[pallet::weight(<T as Config>::WeightInfo::schedule_transfer_task_existing_slot())]`, i got error: 
```
Invalid pallet::call, requires weight attribute i.e. `#[pallet::weight($expr)]`
```

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
justinzhou93: yah, hm that was my thought as well, but when i put it after the ensure, i got this error: 
```the `?` operator can only be applied to values that implement `Try`
the trait `Try` is not implemented for `()`
```
it was much happier when i put it right after the error rather than outside ensure. although i prob need to test it cause it doesn't quite seem right

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
justinzhou93: ahhh i was looking for that type `Parameter`. just had no idea what it was called. i tried `type Balance: Type` and it didn't like it haha. `Parameter` makes it wayyyy happier haha

pallets/automation-time/src/lib.rs
```diff
@@ -245,6 +285,49 @@ pub mod pallet {
...
+			provided_id: Vec<u8>,
+			time: UnixTime,
+			recipient_id: T:: AccountId,
+			amount: T::Balance,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			let recipientExists = account_exists(recipient_id);
+			
+			// check for greater than existential deposit
+			ensure!(amount < T::ExistentialDeposit::get(), Err(<Error<T>>::InvalidAmount)?);
```
justinzhou93, line 321: weird behavior on ensure, see here: https://github.com/OAK-Foundation/OAK-blockchain/pull/74#discussion_r804364321

pallets/automation-time/src/lib.rs
```diff
@@ -245,6 +285,49 @@ pub mod pallet {
...
+		#[pallet::weight(<T as Config>::WeightInfo::schedule_transfer_task_existing_slot())]
+		pub fn schedule_transfer_task(
+			origin: OriginFor<T>,
+			provided_id: Vec<u8>,
+			time: UnixTime,
+			recipient_id: T:: AccountId,
+			amount: T::Balance,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			let recipientExists = account_exists(recipient_id);
```
justinzhou93, line 318: i found `account_exists` in the same file as `ensure_signed` in the substrate repo: https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L948

i'm not sure how `ensure_signed` is getting imported in this file. it looks like it's being set as publicly usable for the crate. `account_exists` is a public function, but does not look like it's available to public use in the crate. so i'm curious if you know a better way to do get if an account exists

pallets/automation-time/src/lib.rs
justinzhou93: also, i'm getting an error on the `<` comparator: 
```
binary operation `<` cannot be applied to type `<T as pallet::Config>::Balance`
the trait `std::cmp::PartialOrd` is not implemented for `<T as pallet::Config>::Balance`
```

Balance is set as u128. u128 should have comparators set. but it seems like it's not quite understanding that - maybe cause it's a generic Config and it's not sure what's getting inserted into the Balance slot. so in this case, is there any alternative to implementing the `std::cmp::PartialOrd` trait? cause i don't really wanna do that haha

justinzhou93: i guess i could just set T::ExistentialDeposit type to u128 rather than Self::Balance. but this seems lazy

pallets/automation-time/src/weights.rs
rhuttman: ohh okay. just make sure to replace it w/ your own benchmark before you merge

pallets/automation-time/src/lib.rs
```diff
@@ -245,6 +300,56 @@ pub mod pallet {
...
+			// ensure!(amount < T::ExistentialDeposit::get(), Err(<Error<T>>::InvalidAmount)?)?;
+			if amount < T::ExistentialDeposit::get() {
+				Err(<Error<T>>::InvalidAmount)?
+			}
+			// check not sent to self
+			if who != recipient_id {
+				Err(<Error<T>>::TransferToSelf)?
+			}
+			let senderBalance = <T as Config>::Currency::free_balance(&who);
+			if senderBalance < amount {
```
justinzhou93, line 345: types don't match. 

```senderBalance: <<T as pallet::Config>::Currency as frame_support::traits::Currency<<T as frame_system::Config>::AccountId>>::Balance;

```amount: <T as pallet::Config>::Balance;

my initial thought was to just adjust our T::Balance to be the same as Currency::Balance: https://github.com/OAK-Foundation/OAK-blockchain/pull/74/files#diff-5d19ac6ca496cd503aa07940ae2dc4a0fdffdfbb6c49726507a70c314135d7a0R80

but that's still not working because it's seemingly not recognizing the type that's being passed in as anything more than T::Balance, since it's not sure what's being injected during launch. so i'm not sure how to coerce the type to like what i'm passing in

File: pallets/automation-time/src/lib.rs (hunk 9 above)
Comment (justinzhou93) [reply pallets/automation-time/src/lib.rs]: RESOLVED: idk why, it disappeared on its own 🤷

File: pallets/automation-time/src/lib.rs (hunk 10 above)
Comment (justinzhou93) [reply pallets/automation-time/src/lib.rs]: RESOLVED: no `account_exists` because accounts without EXISTENTIAL_DEPOSIT will not exist. we want people to be able to send to new accounts

File: pallets/automation-time/src/lib.rs [hunk 12]
```diff
@@ -460,6 +567,25 @@ pub mod pallet {
...
+			
+			// TODO: jzhou fix later
+
+			// <<T as pallet::Config>::Currency as frame_support::traits::Currency<<T as frame_system::Config>::AccountId>>::Balance;
+			// <T as pallet::Config>::Balance;
+			<T as Config>::Currency::transfer(&sender, &recipient, amount, ExistenceRequirement::KeepAlive);
+
+			// let message = Vec::new();
+			// message.push("Sent {amount} tokens from {from} to {to}");
+			// Self::deposit_event(Event::Notify{ message });
```
justinzhou93, line 584: is there an easy way to construct a message if i want to emit an event to say the transfer is complete? or do i need to take each letter and convert into a u8 number and insert into vector?

also, is it fine if i just emit event saying `Transfer complete for : {providedId}`? or does anyone else need more info?

justinzhou93:
image
scheduling def works

justinzhou93:
image

justinzhou93:
image

justinzhou93:
image

justinzhou93:
image

justinzhou93:
image

notify still works with the refactor for more generic transfer function

pallets/automation-time/src/lib.rs
```diff
@@ -228,20 +288,57 @@ pub mod pallet {
...
+		/// * `TransferToSelf`: Sender cannot transfer money to self.
+		/// * `InsufficientFunds`: Amount in sender account is insufficient.
+		/// * `TransferFailed`: Transfer failed for unknown reason.
+		#[pallet::weight(<T as Config>::WeightInfo::schedule_transfer_task_existing_slot())]
+		pub fn schedule_transfer_task(
+			origin: OriginFor<T>,
+			provided_id: Vec<u8>,
+			time: UnixTime,
+			recipient_id: T:: AccountId,
+			#[pallet::compact] amount: BalanceOf<T>,
```
justinzhou93, line 328: i got it to be Compact<u128>, but it's still using the obscenely large number. need to figure out what substrate is doing with the balance pallet. they dump it right into the transfer function without doing anything else, so i'm not sure why it's not working for me: https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L282 

image

pallets/automation-time/src/lib.rs
```diff
@@ -525,6 +653,29 @@ pub mod pallet {
...
+			time: UnixTime,
+		) -> Result<(), Error<T>> {
+			if provided_id.len() == 0 {
+				Err(Error::<T>::EmptyProvidedId)?
+			}
+			Self::is_valid_time(time)?;
+
+			let task_id = Self::schedule_task(who.clone(), provided_id.clone(), time)?;
+			let task: Task<T> = match action {
+				Action::Notify { message } => Task::<T>::create_event_task(who.clone(), provided_id, time, message),
```
justinzhou93, line 669: not sure how to pass in action, cause can't use by reference (cannot pass by reference for task creation). And can't use .clone() for some reason?

pallets/automation-time/src/lib.rs
```diff
@@ -228,20 +288,57 @@ pub mod pallet {
...
+			Self::validate_and_schedule_task(Action::Notify { message }, who.clone(), provided_id, time)?;
+			Ok(().into())
+		}
 
-			let task = Task::<T>::create_event_task(who.clone(), provided_id, time, message);
-			<Tasks<T>>::insert(task_id, task);
+		/// Schedule a task to fire an event with a custom message.
+		///
+		/// Before the task can be scheduled the task must past validation checks.
+		/// * The transaction is signed
```
rhuttman, line 302: you have other checks here

pallets/automation-time/src/lib.rs
```diff
@@ -339,19 +436,23 @@ pub mod pallet {
 			let update_weight = Self::update_task_queue();
 
 			// need to calculate the weight of running just 1 task below.
+			info!("Update Weight: {:?}", update_weight);
```
rhuttman, line 439: old logs

pallets/automation-time/src/lib.rs
```diff
@@ -460,6 +566,28 @@ pub mod pallet {
...
 
+		fn transfer(
+			sender: T::AccountId,
+			recipient: T::AccountId,
+			amount: BalanceOf<T>,
+			task_id: T::Hash,
+		) -> Weight {
+			// check for sufficient funds in the sender Balance
+			let sender_balance = <T as Config>::Currency::free_balance(&sender);
+			if sender_balance < amount {
```
rhuttman, line 577: are we sure the Currency::transfer won't check for this?

pallets/automation-time/src/lib.rs
```diff
@@ -164,6 +195,14 @@ pub mod pallet {
 		TaskDoesNotExist,
 		/// Block time not set.
 		BlockTimeNotSet,
+		/// Amount has to be larger than 0.1 OAK.
+		InvalidAmount,
+		/// Recipient must be a valid address.
+		InvalidRecipient,
```
rhuttman, line 201: you never throw this error

pallets/automation-time/src/lib.rs
```diff
@@ -460,6 +566,28 @@ pub mod pallet {
...
+			amount: BalanceOf<T>,
+			task_id: T::Hash,
+		) -> Weight {
+			// check for sufficient funds in the sender Balance
+			let sender_balance = <T as Config>::Currency::free_balance(&sender);
+			if sender_balance < amount {
+				Self::deposit_event(Event::InsufficientFunds { task_id });
+				return 10_000;
+			}
+			match <T as Config>::Currency::transfer(&sender, &recipient, amount, ExistenceRequirement::KeepAlive) {
```
rhuttman, line 581: since you only care about the error you should use an if let for this

pallets/automation-time/src/lib.rs
rhuttman: you can just create the struct here, ie dont use the helpers. I really just needed the helper for the tests. I couldn;t figure out how else to make the tasks in the. test file.

pallets/automation-time/src/mock.rs
```diff
@@ -16,7 +16,8 @@
 // limitations under the License.
 
 use crate as pallet_automation_time;
-use frame_support::{parameter_types, traits::Everything, weights::Weight};
+// use crate as pallet_balances;
```
rhuttman, line 19: i think you meant to delete this

pallets/automation-time/src/mock.rs
```diff
@@ -43,10 +45,30 @@ frame_support::construct_runtime!(
...
+parameter_types! {
+	pub const MaxLocks: u32 = 50;
+	pub const MaxReserves: u32 = 50;
+}
+
+impl pallet_balances::Config for Test {
+	type MaxLocks = MaxLocks;
+	/// The type for recording an account's balance.
+	type Balance = Balance;
+	/// The ubiquitous event type.
```
rhuttman, line 62: you can remove these comments for the mock

pallets/automation-time/src/benchmarking.rs
```diff
@@ -76,6 +76,26 @@ benchmarks! {
...
 	}: schedule_notify_task(RawOrigin::Signed(caller), vec![10], time, vec![4, 5])
 
+	// First transfer task for a slot
+	schedule_transfer_task_new_slot {
+		let caller: T::AccountId = whitelisted_caller();
+		let recipient: T::AccountId = whitelisted_caller();
+		let time: u64 = 120;
+		let count: u8 = 1;
+
+		let task_id: T::Hash = schedule_tasks::<T>(caller.clone(), time, 1);
```
rhuttman, line 86: since this is for a new slot you shouldn't schedule. Also since we know the existing slot takes longer we really only need 1 benchmark for the existing. I need to clean up this file to just one have per extrinsic (the bigger one)

pallets/automation-time/src/tests.rs
```diff
@@ -460,6 +526,24 @@ fn add_task_to_task_queue(
 	task_id
 }
 
+fn _add_transfer_task_to_task_queue(
```
rhuttman, line 529: _ at the start

rhuttman: also seems like you can combine most of this and add_notify_task_to_task_queue

pallets/automation-time/src/weights.rs
```diff
@@ -62,6 +64,24 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
...
+	// Storage: AutomationTime Tasks (r:1 w:1)
+	// Storage: AutomationTime ScheduledTasks (r:1 w:1)
+	// TODO: this is placeholder, jzhou will fix
```
rhuttman, line 70: just run the command at the top of this file but change the name of the file it outputs to. Then grab the results you want and throw them in here. The command outputs essentially this file minus how i expose the results

rhuttman (changes requested):
I don't see a test that your transfer task works when run

pallets/automation-time/src/lib.rs
justinzhou93: ahh i changed this to be this instead. fixes my return issue
```suggestion
			match <T as Config>::Currency::transfer(&sender, &recipient, amount, ExistenceRequirement::KeepAlive) {
				Ok(_number)  => Self::deposit_event(Event::SuccesfullyTransferredFunds { task_id }),
        		        Err(_e) => Self::deposit_event(Event::TransferFailed { task_id }),
			};
```

pallets/automation-time/src/lib.rs
justinzhou93: image
it will. but still, the problem is that it gives a generic `automationTime.TransferFailed` error. although maybe we're ok with that? i wanted to give a more specific `InsufficientFunds` error

justinzhou93: is the additional clarity worth the calculation time?

pallets/automation-time/src/tests.rs
justinzhou93: just removing for now, since we don't need

pallets/automation-time/src/lib.rs
```diff
@@ -41,27 +41,37 @@ mod benchmarking;
 pub mod weights;
 
 use core::convert::TryInto;
-use frame_support::{inherent::Vec, pallet_prelude::*, sp_runtime::traits::Hash, BoundedVec};
+use frame_support::{inherent::Vec, pallet_prelude::*, sp_runtime::traits::Hash, BoundedVec, traits::{ Currency }};
```
justinzhou93, line 44: remove

pallets/automation-time/src/lib.rs
```diff
@@ -460,6 +553,25 @@ pub mod pallet {
 			10_000
 		}
 
+		fn transfer(
+			sender: T::AccountId,
+			recipient: T::AccountId,
+			amount: BalanceOf<T>,
+			task_id: T::Hash,
+		) -> Weight {
+			// let sender_balance = <T as Config>::Currency::free_balance(&sender);
```
justinzhou93, line 562: remove

pallets/automation-time/src/lib.rs
```diff
@@ -460,6 +553,25 @@ pub mod pallet {
...
+			// if sender_balance < amount {
+			// 	Self::deposit_event(Event::InsufficientFunds { task_id });
+			// 	return 10_000;
+			// }
+			match <T as Config>::Currency::transfer(&sender, &recipient, amount, ExistenceRequirement::KeepAlive) {
+				Ok(_number)  => Self::deposit_event(Event::SuccesfullyTransferredFunds { task_id }),
+        		Err(_e) => Self::deposit_event(Event::TransferFailed { task_id }),
```
justinzhou93, line 569: maybe do match here, or surface error

pallets/automation-time/src/mock.rs
```diff
@@ -16,7 +16,7 @@
 // limitations under the License.
 
 use crate as pallet_automation_time;
-use frame_support::{parameter_types, traits::Everything, weights::Weight};
+use frame_support::{parameter_types, traits::{Everything}, weights::Weight};
```
justinzhou93, line 19: remove brackets

pallets/automation-time/src/mock.rs
```diff
@@ -103,6 +123,12 @@ impl<Test: frame_system::Config> pallet_automation_time::WeightInfo for MockWeig
 	fn schedule_notify_task_existing_slot() -> Weight {
 		0
 	}
+	fn schedule_transfer_task_new_slot() -> Weight {
```
justinzhou93, line 126: remove

justinzhou93:
image
what the heck. balances.transfer does the same thing

pallets/automation-time/src/lib.rs
```diff
@@ -460,6 +546,25 @@ pub mod pallet {
...
+			task_id: T::Hash,
+		) -> Weight {
+			match <T as Config>::Currency::transfer(
+				&sender,
+				&recipient,
+				amount,
+				ExistenceRequirement::KeepAlive,
+			) {
+				Ok(_number) => Self::deposit_event(Event::SuccesfullyTransferredFunds { task_id }),
+				Err(e) => Self::deposit_event(Event::TransferFailed { task_id, error: e }),
```
justinzhou93, line 562: acceptably pretty error? 
image

my only concern here is: is there anything that's going to be "PII" that's being printed? i don't think so, since blockchain is all public anyway, but just curious

pallets/automation-time/src/lib.rs
justinzhou93: note: balance pallet transfer def uses this super large number too

pallets/automation-time/src/weights.rs
justinzhou93: TODO: need benchmarking code update

pallets/automation-time/src/lib.rs
justinzhou93: surface error, as seen in picture/post below

pallets/automation-time/src/weights.rs
justinzhou93: also FYI, transfer is quite a bit heavier, as seen here. i copied over a task that was doing 90_000 and i couldn't accomplish without 120_000. https://github.com/OAK-Foundation/OAK-blockchain/pull/74/files#diff-8aa2921b8ffd2c8a5901003639e03bb443a2032d62c687606ea6a9f7bbcaf99cR554

justinzhou93:
some remaining things to note:
1. balances.transfer uses the obscenely large number too: https://github.com/OAK-Foundation/OAK-blockchain/pull/74#discussion_r807332572
2. transfer is a good bit heavier than notify, as seen in the test here: https://github.com/OAK-Foundation/OAK-blockchain/pull/74/files#r807360103
3. worth checking note on whether the error being passed up on TransferFailed is acceptably pretty/PII free: https://github.com/OAK-Foundation/OAK-blockchain/pull/74#discussion_r807330499

pallets/automation-time/src/lib.rs
```diff
@@ -237,20 +281,63 @@ pub mod pallet {
...
-			}
 			if message.len() == 0 {
 				Err(Error::<T>::EmptyMessage)?
 			}
-			Self::is_valid_time(time)?;
 
-			let task_id = Self::schedule_task(who.clone(), provided_id.clone(), time)?;
+			Self::validate_and_schedule_task(
+				Action::Notify { message },
+				who.clone(),
```
rhuttman, line 290: you don't need to clone here. I also did that

pallets/automation-time/src/lib.rs
rhuttman: well these errors are what the Balance transfer API would return right?

pallets/automation-time/src/lib.rs
```diff
@@ -482,6 +567,8 @@ pub mod pallet {
 					Some(task) => {
 						let action_weight = match task.action {
 							Action::Notify { message } => Self::run_notify_task(message),
+							Action::Transfer { sender, recipient, amount } =>
+								Self::run_transfer_task(sender, recipient, amount, task_id.clone()),
```
rhuttman, line 571: do we need to clone task_id here?

rhuttman (changes requested):

pallets/automation-time/src/lib.rs
justinzhou93: yah we do i think. i think when you do .iter(), it references the individual task_ids from the vector as reference, but i’m passing in an actual task_id for the run_transfer_task function. unless we want to always pass in reference for run_transfer_task function?

rhuttman (approved):

PR #78, opened 2022-02-16
Limit how far out tasks can be scheduled

rhuttman:


pallets/automation-time/src/lib.rs
```diff
@@ -411,6 +420,10 @@ pub mod pallet {
 				Err(<Error<T>>::PastTime)?;
 			}
 
+			if scheduled_time >= current_time_slot + T::MaxScheduleSeconds::get() {
```
justinzhou93, line 423: just greater than rather than >=, i think? at least according to this comment: /// - Not be more than MaxScheduleSeconds out

pallets/automation-time/src/lib.rs
```diff
@@ -177,6 +181,8 @@ pub mod pallet {
 		InvalidTime,
 		/// Time must be in the future.
 		PastTime,
+		/// Time cannot be too far in the future.
```
justinzhou93, line 184: do we need to explain why here for documentation purposes? and do we need to tell them exactly how far out in the UI?

justinzhou93 (approved):
2 notes, but for the most part, lgtm

pallets/automation-time/src/lib.rs
rhuttman: There is no real why. We just want a limit. How far out is based on environment. Will need to be in our API docs

justinzhou93 (approved):
re

PR #80, opened 2022-02-17
Fixing Automation time benchmarks

rhuttman:


pallets/automation-time/src/lib.rs
```diff
@@ -309,8 +309,8 @@ pub mod pallet {
 		/// * `TimeSlotFull`: Time slot is full. No more tasks can be scheduled for this time.
 		/// * `InvalidAmount`: Amount has to be larger than 0.1 OAK.
 		/// * `TransferToSelf`: Sender cannot transfer money to self.
-		/// * `TransferFailed`: Transfer failed.
-		#[pallet::weight(<T as Config>::WeightInfo::schedule_transfer_task_existing_slot())]
+		/// * `TransferFailed`: Transfer failed for unknown reason.
```
justinzhou93, line 312: why add back "for unknown reason"? the reason should be returned along with the error from the transfer function from balances pallet, along with explanation that gets bubbled up as an event.

justinzhou93 (approved):
1 nit otherwise lgtm!

pallets/automation-time/src/lib.rs
rhuttman: was a bad merge. I will fix it when I do a round of code cleanup this weekend

PR #81, opened 2022-02-17
Valve can stop scheduled tasks

rhuttman:


irsal (approved):

pallets/valve/src/lib.rs
```diff
@@ -169,6 +178,38 @@ pub mod pallet {
 			};
 			Ok(())
 		}
+
+		/// Stop all scheduled tasks from running.
+		#[pallet::weight(T::DbWeight::get().read * T::DbWeight::get().write)]
+		pub fn stop_scheduled_tasks(origin: OriginFor<T>) -> DispatchResult {
```
justinzhou93, line 184: it seems like these 2 functions aren't being used anywhere. i'm just assuming it's for a future PR?

justinzhou93 (approved):

pallets/valve/src/lib.rs
rhuttman: They are extrinsics

PR #82, opened 2022-02-18
NEU genesis allocation

rhuttman:


node/src/chain_spec.rs
```diff
@@ -294,3 +294,137 @@ fn testnet_genesis(
...
+	properties.insert("ss58Format".into(), SS_58_FORMAT.into());
+
+	ChainSpec::from_genesis(
+		// Name
+		"Neumann Network",
+		// ID
+		"neumann",
+		ChainType::Live,
+		move || {
+			let allocation_json = &include_bytes!("../../distribution/neumann_alloc.json")[..];
```
irsal, line 312: does this get prioritized over the 3 big accounts? ie will each of the 3 accounts below (yours, mine and chris') have < 333.33MM NEU?

distribution/neumann_alloc.json
```diff
@@ -0,0 +1,18 @@
...
+	[
+		"5CM2JyPHnbs81Cu8GzbraqHiwjeNwX3c9Rr5nXkJfwK9fwrk",
+		3333333333333333333
+	],
+    [
+		"5GcD1vPdWzBd3VPTPgVFWL9K7b27A2tPYcVTJoGwKcLjdG5w",
+        3333333333333333333
+	],
+    [
+        "5DhqAmCUoY7FiMCY7YpTyGT4fFpdXYmYjzzE9fHJwLDyBafZ",
```
justinzhou93, line 15: how did you split this into 4 accounts with these splits? i can't see how you're inserting the ratios in the chainspec.rs file below

node/src/chain_spec.rs
rhuttman: I am going to take the output from subquery and transform it into the allocation json. Our accounts will be in there with whatever amount we have now. We only use this json for endowed accounts

distribution/neumann_alloc.json
rhuttman: I made them up as tests. The real one will come later

justinzhou93: ah yah totally, i just meant i can't seem to find where you defined the ratios in the code

justinzhou93 (approved):
code/results lgtm, but i'll def defer to irsal on whether this works, since she's doing some testing

irsal (approved):
💪

PR #83, opened 2022-02-18
fix schedule_transfer_task to be schedule_native_transfer_task

justinzhou93:


justinzhou93:
didn't spin this change up locally on the blockchain, just did local tests, cause seems like it should be pretty easy. can def do if needed tho!

rhuttman (approved):

rhuttman (approved):

PR #88, opened 2022-02-28
Vesting pallet

rhuttman:


rhuttman:
This will be followed by a PR to
- Add benchmarks
- Add to runtimes
- Add to chain specs

pallets/vesting/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+			if now == 0 {
+				Err(Error::<T>::BlockTimeNotSet)?
+			}
+			let now = now / 1000;
+			let diff_to_min = now % 3600;
+			Ok(now - diff_to_min)
+		}
+
+		/// Mint tokens for any accounts that have vested.
+		pub fn vest() {
```
justinzhou93, line 111: when does this get called? and should we test in the test file?

justinzhou93 (approved):

PR #95, opened 2022-03-02
Execution fee

rhuttman:


pallets/automation-time/src/exchange.rs
```diff
@@ -0,0 +1,143 @@
...
+///
+/// The unbalance handler is given 2 unbalanceds in [`OnUnbalanced::on_unbalanceds`]: fee and
+/// then tip.
+impl<T, C, OU> NativeTokenExchange<T> for CurrencyAdapter<C, OU>
+where
+	T: Config,
+	C: Currency<<T as frame_system::Config>::AccountId>,
+	C::PositiveImbalance: Imbalance<
+		<C as Currency<<T as frame_system::Config>::AccountId>>::Balance,
+		Opposite = C::NegativeImbalance,
```
justinzhou93, line 85: what is this syntax where you set Opposite equal to something inside of a generic? and what is Opposite mean? i didn't see it imported (i get the gist of what it's trying to do, but idk how it works syntactically)

pallets/automation-time/src/exchange.rs
```diff
@@ -0,0 +1,143 @@
...
+		dest: &T::AccountId,
+		value: Self::Balance,
+	) -> Result<(), DispatchError> {
+		C::transfer(source, dest, value, ExistenceRequirement::KeepAlive)?;
+		Ok(())
+	}
+
+	// Ensure the fee can be paid.
+	fn can_pay_fee(who: &T::AccountId, fee: Self::Balance) -> Result<(), DispatchError> {
+		if fee.is_zero() {
```
justinzhou93, line 115: just ot be sure, this is saying that htere's a fee, not that the fee can be paid, right (contradicts the comment above function)? if it is saying that the fee cannot be paid, why return ok? shouldn't it return error? and if it's fee is zero, then when will fee be zero? don't we want ot stop DDOS by always having fee?

pallets/automation-time/src/lib.rs
```diff
@@ -769,5 +781,14 @@ pub mod pallet {
...
 			T::Hashing::hash_of(&task_hash_input)
 		}
+
+		fn calculate_execution_fee(action: &Action<T>) -> BalanceOf<T> {
+			let raw_weight = match action {
+				Action::Notify { message: _ } => 1_000u32,
+				Action::NativeTransfer { sender: _, recipient: _, amount: _ } => 2_000u32,
+			};
+			let weight = <BalanceOf<T>>::from(raw_weight);
+			T::ExecutionWeightFee::get().saturating_mul(weight)
```
justinzhou93, line 791: why is execution weight fee static? shouldn't it scale with size?

pallets/automation-time/src/exchange.rs
```diff
@@ -0,0 +1,143 @@
...
+		source: &T::AccountId,
+		dest: &T::AccountId,
+		value: Self::Balance,
+	) -> Result<(), DispatchError> {
+		C::transfer(source, dest, value, ExistenceRequirement::KeepAlive)?;
+		Ok(())
+	}
+
+	// Ensure the fee can be paid.
+	fn can_pay_fee(who: &T::AccountId, fee: Self::Balance) -> Result<(), DispatchError> {
```
justinzhou93, line 114: should fee calcs be in this folder as well or will that be separate pallet? imo native token will ahve native token fees, so we should put in same folder, right?

pallets/automation-time/src/exchange.rs
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	///Transfer some liquid free balance to another staker.
+	/// This is a very high-level function. It will ensure all appropriate fees are paid and no imbalance in the system remains.
+	fn transfer(
+		source: &T::AccountId,
+		dest: &T::AccountId,
+		value: Self::Balance,
+	) -> Result<(), DispatchError> {
+		C::transfer(source, dest, value, ExistenceRequirement::KeepAlive)?;
```
justinzhou93, line 109: should we do this check ability to pay prior to transfer so that we don't complete transfer without knowing if they can pay fees after (https://github.com/OAK-Foundation/OAK-blockchain/pull/95/files#diff-4b05c2f8a100ed7a26a7aad41276bfddd68dda106b04fde6ac6a2f42ac4ad94dR114)?

i.e. they have 1 OAK in pocket. they transfer 1 OAK. no money left for fees

justinzhou93 (approved):
lgtm, just a couple of questions around how fees are gonna work. realized i'm a little fuzzy on that still

pallets/automation-time/src/exchange.rs
rhuttman: The execution fee will be calculated differently for each automation pallet

justinzhou93 (approved):

pallets/automation-time/src/exchange.rs
```diff
@@ -0,0 +1,143 @@
...
+		+ Debug
+		+ Default
+		+ scale_info::TypeInfo;
+	type LiquidityInfo: Default;
+
+	/// The minimum balance any single account may have. This is equivalent to the `Balances`
+	/// module's `ExistentialDeposit`.
+	fn minimum_balance() -> Self::Balance;
+
+	///Transfer some liquid free balance to another staker.
```
irsal, line 52: staker?

pallets/automation-time/src/exchange.rs
```diff
@@ -0,0 +1,143 @@
+// This file is part of OAK Blockchain.
```
irsal, line 1: pls explain name of file `exchange.rs`

pallets/automation-time/src/lib.rs
```diff
@@ -208,6 +211,10 @@ pub mod pallet {
 		InvalidAmount,
 		/// Sender cannot transfer money to self.
 		TransferToSelf,
+		/// Insufficient balance to pay execution fee.
+		InsufficientBalance,
+		/// Account liquidity restrictions prevent withdrawal.
+		LiquidityRestrictions,
```
irsal, line 217: we don't allow for liquidity in our chain tho?

pallets/automation-time/src/lib.rs
irsal: can someone explain the resolution to this pls?

runtime/neumann/src/lib.rs
```diff
@@ -723,9 +723,25 @@ parameter_types! {
...
+}
+pub struct DealWithExecutionFees<R>(sp_std::marker::PhantomData<R>);
+impl<R> OnUnbalanced<NegativeImbalance<R>> for DealWithExecutionFees<R>
+where
+	R: pallet_balances::Config + pallet_treasury::Config,
+	pallet_treasury::Pallet<R>: OnUnbalanced<NegativeImbalance<R>>,
+{
+	fn on_unbalanceds<B>(mut fees: impl Iterator<Item = NegativeImbalance<R>>) {
+		if let Some(fees) = fees.next() {
+			// 80% burned, 20% to the treasury
```
irsal, line 736: will discuss/confirm this with sean within the hour

runtime/turing/src/lib.rs
```diff
@@ -721,9 +723,26 @@ parameter_types! {
 	pub const MaxBlockWeight: Weight = MAXIMUM_BLOCK_WEIGHT;
 	pub const MaxWeightPercentage: Perbill = SCHEDULED_TASKS_INITIALIZE_RATIO;
 	pub const SecondsPerBlock: u64 = MILLISECS_PER_BLOCK / 1000;
+	pub const ExecutionWeightFee: Balance = 12;
+}
+
+pub struct DealWithExecutionFees<R>(sp_std::marker::PhantomData<R>);
```
irsal, line 729: gosh, the repeated code kills me b/w the 2 runtimes

irsal (commented):
❔ ❔ ❔ ❔ ❓

runtime/turing/src/lib.rs
rhuttman: ya I need to do more work to figure out how import more thing from a common runtime

pallets/automation-time/src/lib.rs
rhuttman: This is the error the Balance pallet will throw. I am just matching it.

The error occurs because you have enough funds, but the value of your funds after the action would be below what the withdrawal type allows for.

pallets/automation-time/src/exchange.rs
rhuttman: I am bad at names....

The point of that file is to hold the trait we want our automation pallets to use to interact with Balances. This includes fees and transfers. At some point I will extract this into another util lib so it can used between pallets.

pallets/automation-time/src/exchange.rs
rhuttman: Ya i just copied the documentation from the Balance pallet. I was debating between making it clearer to me vs making this is a straight pass through

irsal (approved):
thx for addressing

PR #96, opened 2022-03-02
0.9.17 upgrade

rhuttman:


justinzhou93 (approved):

runtime/turing/src/lib.rs
```diff
@@ -573,6 +573,8 @@ impl cumulus_pallet_xcmp_queue::Config for Runtime {
 	type ChannelInfo = ParachainSystem;
 	type VersionWrapper = ();
 	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
+	type ControllerOrigin = EnsureRoot<AccountId>;
```
irsal, line 576: this is the controller account for sudo?

irsal (approved):

runtime/turing/src/lib.rs
rhuttman: yes

PR #97, opened 2022-03-02
Valve weights

rhuttman:


node/src/chain_spec/neumann.rs
```diff
@@ -5,14 +5,14 @@ use sc_service::ChainType;
 use sc_telemetry::TelemetryEndpoints;
 use sp_core::{crypto::UncheckedInto, sr25519};
 
+use super::TELEMETRY_URL;
```
irsal, line 8: thx for the lint

pallets/valve/src/benchmarking.rs
```diff
@@ -0,0 +1,63 @@
+// This file is part of OAK Blockchain.
+
+// Copyright (C) 2021 OAK Network
```
irsal, line 3: 2022

pallets/valve/src/weights.rs
```diff
@@ -0,0 +1,105 @@
+
+//! Autogenerated weights for `pallet_valve`
+//!
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
+//! DATE: 2022-03-02, STEPS: `20`, REPEAT: 64, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
irsal, line 5: how frequently do we have to regenerate these weights?

irsal (approved):

irsal:
merge conf, probs need to update 0.9.17 client as well

pallets/valve/src/weights.rs
rhuttman: Whenever we change something that could change the weights. That could be changing the logic in a function or changing pallet config params

pallets/valve/src/benchmarking.rs
```diff
@@ -0,0 +1,63 @@
...
+	}: close_pallet_gate(RawOrigin::Root, b"System".to_vec())
+
+
+	open_pallet_gate {
+		let pallet_name = b"System".to_vec();
+		ClosedPallets::<T>::insert(pallet_name.clone(), ());
+	}: _(RawOrigin::Root, pallet_name)
+
+	open_pallet_gates {
+		ClosedPallets::<T>::insert(b"System".to_vec(), ());
```
justinzhou93, line 48: if i remember correctly, you mentioned there are WAYYY more gates than this. shouldn't we open all of the gates that exist to make sure we get the correct amount? or alternatively use iterator `v`?

justinzhou93 (approved):
1 question, otherwise lgtm

PR #99, opened 2022-03-03
Safe Math

rhuttman:


irsal (approved):
ez, is there a source / best practices we can read about this?

rhuttman:
@irsal sadly I cannot find such a thing. By using these functions we ensure errors are never thrown due to number being too large or small. Since we can;t afford our code to panic it's just safer to always use these

justinzhou93 (approved):
lgtm. curious on 1 point. ik saturating just safely exits and all that, but is there ever a world where we want to throw an error before we exit (just so we know what happened in that block if we're debugging things that don't get finished)? or is that not possible without any weight?

justinzhou93 (approved):

PR #100, opened 2022-03-07
Release prep for 1.2.7 && 276 runtime 

irsal:


runtime/neumann/src/lib.rs
```diff
@@ -176,10 +176,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("neumann"),
 	impl_name: create_runtime_str!("neumann"),
 	authoring_version: 1,
-	spec_version: 275,
+	spec_version: 276,
 	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 3,
+	transaction_version: 4,
```
irsal, line 182: fyi @rhuttman - i'm 90% sure this is what did the trick to kick off the runtime upgrade

rhuttman (approved):

runtime/neumann/src/lib.rs
rhuttman: ya adding this did it: https://github.com/OAK-Foundation/OAK-blockchain/blob/master/pallets/valve/src/lib.rs#L213

PR #113, opened 2022-03-14
Change wallet address technical council member

irsal:


justinzhou93 (approved):
i'm guessing this is laura's technical committee address? if so, would like her to sign off too haha, but lgtm on my end!

image

PR #124, opened 2022-04-04
0.9.18

rhuttman:


node/src/command.rs
```diff
@@ -326,6 +326,7 @@ pub fn run() -> Result<()> {
 		// Fix as this run the node
 		None => {
 			let runner = cli.create_runner(&cli.run.normalize())?;
+			let collator_options = cli.run.collator_options();
```
justinzhou93, line 329: i'm assuming there are defaults that come out of this function? cause it doesn't seem like we're adding any specific options here

justinzhou93 (approved):
i forget, what's the command for formatting the codebase again? mind writing that down in README or doc? i always forget what it is and end up not formatting.

node/src/command.rs
rhuttman: `	/// Create [`CollatorOptions`] representing options only relevant to parachain collator nodes
	pub fn collator_options(&self) -> CollatorOptions {
		CollatorOptions { relay_chain_rpc_url: self.relay_chain_rpc_url.clone() }
	}`

irsal (commented):
- Add change log
- Does this change bump any of the runtime versions?

irsal (approved):

PR #127, opened 2022-04-12
swap min to hour

justinzhou93:


justinzhou93:
@lreesby should be a good base to work off of. should be mostly done, unit tests and benchmarks and all. i will run QA testing tomorrow morning, but otherwise, should be good to go

justinzhou93:
QA'ed following scenarios:
1. migration removes all existing tasks from all queues and maps
2. storage version bumps to version 2 (storage version exists separately for each pallet)
3. last time slot is set to the current hour slot for both last time slot and last missed time slot, before it continues forward. similarly to before, the last missed time slot lags by 1 slot, so in this case, 3600 seconds.
4. errors when don't schedule on the hour
5. 60 transfers per block, but will complete many over the course of many blocks throughout the hour (max set to 1000 right now)

pallets/automation-time/src/migrations.rs
```diff
@@ -23,3 +23,41 @@ pub mod v1 {
...
+	use crate::{LastTimeSlot, Pallet, Task, Vec, TaskQueue, MissedQueue};
+
+	use super::*;
+
+	pub fn migrate<T: Config>() -> Weight {
+		info!(target: "automation-time", "Migrating automation-time v2");
+		let pallet_prefix: &[u8] = b"AutomationTime";
+
+		let scheduled_tasks_prefix: &[u8] = b"ScheduledTasks";
+		let _scheduled_tasks: Vec<_> = storage_key_iter::<u64, BoundedVec<T::Hash, T::MaxTasksPerSlot>, Twox64Concat>(pallet_prefix, scheduled_tasks_prefix)
```
rhuttman, line 39: how does this work

pallets/automation-time/src/tests.rs
```diff
@@ -785,6 +785,49 @@ fn on_init_shutdown() {
...
 }
 
+#[test]
+fn migration_v1() {
+	new_test_ext(START_BLOCK_TIME).execute_with(|| {
+		AutomationTime::on_initialize(1);
+		let new_timestamp = START_BLOCK_TIME + (3600 * 1_000);
+		Timestamp::set_timestamp(new_timestamp);
+		v1::migrate::<Test>();
+		if let Some((updated_last_time_slot, updated_last_missed_slot)) =
```
rhuttman, line 795: you don't have a fail in there for the else

pallets/automation-time/src/tests.rs
```diff
@@ -785,6 +785,49 @@ fn on_init_shutdown() {
...
+
+		v2::migrate::<Test>();
+		Timestamp::set_timestamp(START_BLOCK_TIME + (3600 * 1_000));
+		if let Some((updated_last_time_slot, updated_last_missed_slot)) =
+			AutomationTime::get_last_slot()
+		{
+			assert_eq!(updated_last_time_slot, LAST_BLOCK_TIME,);
+			assert_eq!(updated_last_missed_slot, LAST_BLOCK_TIME,);
+		}
+		match AutomationTime::get_scheduled_tasks(SCHEDULED_TIME) {
```
rhuttman, line 820: also need to check the Tasks map

rhuttman (changes requested):

pallets/automation-time/src/migrations.rs
justinzhou93: the .drain() is a function on an iterator, which removes everything from the storage map  https://docs.substrate.io/rustdocs//latest/src/frame_support/storage/migration.rs.html#61

and storage_key_iter just provides the iterator to drain the scheduledTasks storage: https://docs.substrate.io/rustdocs//latest/src/frame_support/storage/migration.rs.html#193

was following moonbeam here: https://github.com/PureStake/moonbeam/blob/e7b414a1370f7af5705e31835336460181be9267/pallets/asset-manager/src/migrations.rs#L82

pallets/automation-time/src/tests.rs
justinzhou93: gahh i knew i forgot something on this test

runtime/neumann/src/lib.rs
```diff
@@ -212,7 +212,7 @@ pub const fn deposit(items: u32, bytes: u32) -> Balance {
 pub const EXISTENTIAL_DEPOSIT: Balance = DOLLAR / 10;
 
 /// We use at most 10% of the block weight running scheduled tasks during `on_initialize`.
-const SCHEDULED_TASKS_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
+const SCHEDULED_TASKS_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
```
rhuttman, line 215: also do for TUR

rhuttman (changes requested):

rhuttman (approved):
lgtm

rhuttman (approved):
lgtm

PR #128, opened 2022-04-13
Recurring tasks support

lreesby:
TODO:
- [x] wrap scheduling in transaction
- [x] fix errors for 'get_current_time_slot' calls
- [x] fees
- [x] tests
- [x] benchmarking


pallets/automation-time/src/lib.rs
```diff
@@ -764,33 +772,58 @@ pub mod pallet {
...
-						if task_ids[i] == task_id {
-							if task_ids.len() == 1 {
-								<ScheduledTasks<T>>::remove(task.time);
-							} else {
-								task_ids.remove(i);
-								<ScheduledTasks<T>>::insert(task.time, task_ids);
+			let mut times: Vec<UnixTime> = task.time.to_vec();
+
+			times.sort_unstable();
+			times.dedup();
```
justinzhou93, line 797: could prob throw these 3 lines (to_vec up until dedup) into separate function

pallets/automation-time/src/lib.rs
```diff
@@ -764,33 +772,58 @@ pub mod pallet {
...
+			scheduled_times.sort_unstable();
+			scheduled_times.dedup();
+
+			return *scheduled_times.last().unwrap() <= current_time_slot;
+		}
+
 		fn remove_task(task_id: T::Hash, task: Task<T>) {
+			let current_time_slot = match Self::get_current_time_slot() {
+				Ok(time_slot) => time_slot,
+				Err(_) => return, // fix error
```
justinzhou93, line 791: reminder to fix error!

pallets/automation-time/src/lib.rs
```diff
@@ -764,33 +772,58 @@ pub mod pallet {
...
-								<ScheduledTasks<T>>::remove(task.time);
-							} else {
-								task_ids.remove(i);
-								<ScheduledTasks<T>>::insert(task.time, task_ids);
+			let mut times: Vec<UnixTime> = task.time.to_vec();
+
+			times.sort_unstable();
+			times.dedup();
+
+			for scheduled_time in times.iter().rev() {
```
justinzhou93, line 799: i'm assuming rev is because sorted the opposite way?

pallets/automation-time/src/lib.rs
```diff
@@ -764,33 +772,58 @@ pub mod pallet {
...
+				if *scheduled_time < current_time_slot { break; }
+				match Self::get_scheduled_tasks(*scheduled_time) {
```
justinzhou93, line 801: not sure i understand this match statement. i get that it removes from scheduled tasks if scheduled time is found there. but why only remove from task queue if scheduled time is not found?

pallets/automation-time/src/lib.rs
```diff
@@ -801,30 +834,32 @@ pub mod pallet {
 			Self::deposit_event(Event::TaskCancelled { who: task.owner_id, task_id });
 		}
 
-		/// Schedule task and return it's task_id.
+		/// Schedule tasks and return it's task_id.
 		pub fn schedule_task(
 			owner_id: AccountOf<T>,
 			provided_id: Vec<u8>,
-			time: u64,
+			time: Vec<UnixTime>,
```
justinzhou93, line 841: prob change the variable name to plural

pallets/automation-time/src/lib.rs
```diff
@@ -764,33 +772,58 @@ pub mod pallet {
...
-				},
-				Some(mut task_ids) =>
-					for i in 0..task_ids.len() {
-						if task_ids[i] == task_id {
-							if task_ids.len() == 1 {
-								<ScheduledTasks<T>>::remove(task.time);
-							} else {
-								task_ids.remove(i);
-								<ScheduledTasks<T>>::insert(task.time, task_ids);
+			let mut times: Vec<UnixTime> = task.time.to_vec();
```
justinzhou93, line 794: wonder if we change task.time to plural

pallets/automation-time/src/lib.rs
```diff
@@ -801,30 +834,32 @@ pub mod pallet {
...
-					<ScheduledTasks<T>>::insert(time, task_ids);
-				},
-				Some(mut task_ids) => {
-					if let Err(_) = task_ids.try_push(task_id) {
-						Err(Error::<T>::TimeSlotFull)?
-					}
-					<ScheduledTasks<T>>::insert(time, task_ids);
-				},
+			for scheduled_time in time.iter() {
+				match Self::get_scheduled_tasks(*scheduled_time) {
```
justinzhou93, line 850: i forget if deferenced pointer clones the value for ya before it mutates it. prob good to unit test, but otherwise, this works for me!

pallets/automation-time/src/lib.rs
```diff
@@ -835,19 +870,29 @@ pub mod pallet {
...
+				Self::is_valid_time(*scheduled_time)?;
+			}
 
 			let fee = Self::calculate_execution_fee(&action);
 			T::NativeTokenExchange::can_pay_fee(&who, fee.clone())
 				.map_err(|_| Error::InsufficientBalance)?;
 
-			let task_id = Self::schedule_task(who.clone(), provided_id.clone(), time)?;
-			let task: Task<T> = Task::<T> { owner_id: who.clone(), provided_id, time, action };
+			let task_id = Self::schedule_task(who.clone(), provided_id.clone(), cleaned_times.clone())?;
```
justinzhou93, line 894: huh i didn't know you could clone vectors. double check that works. but otherwise, might need to do something else here.

justinzhou93 (changes requested):
good first cut! couple of q's + notes

pallets/automation-time/src/lib.rs
```diff
@@ -764,33 +772,58 @@ pub mod pallet {
...
 		}
 
+		fn is_completed_task(time: BoundedVec<UnixTime, T::MaxRecurringTimes>) -> bool {
+			let current_time_slot = match Self::get_current_time_slot() {
+				Ok(time_slot) => time_slot,
+				Err(_) => return false, // fix error
+			};
+
+			let mut scheduled_times = time.to_vec();
+			scheduled_times.sort_unstable();
```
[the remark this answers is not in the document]
lreesby: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_unstable

"This sort is unstable (i.e., may reorder equal elements), in-place (i.e., does not allocate), and O(n * log(n)) worst-case."

pallets/automation-time/src/lib.rs
lreesby: it's sorted in ascending order but I want traverse in descending order.  so `rev`.

pallets/automation-time/src/lib.rs
lreesby: will definitely double check.  it seems to support it from the docs.  https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Clone

pallets/automation-time/src/lib.rs
lreesby: this made sense when you could only schedule for one time, so either your task is in `ScheduledTasks` or `TaskQueue`.  I'll need to update this function because now you could have tasks in both places at once.

runtime/turing/src/lib.rs
```diff
@@ -887,6 +887,7 @@ impl pallet_democracy::Config for Runtime {
 
 parameter_types! {
 	pub const MaxTasksPerSlot: u32 = 2;
+	pub const MaxRecurringTimes: u32 = 24;
```
justinzhou93, line 890: you gotta insert this into the mock file to make sure tests pass

pallets/automation-time/src/lib.rs
```diff
@@ -787,52 +798,62 @@ pub mod pallet {
...
-								}
+			if let Some((last_time_slot, _)) = Self::get_last_slot() {
+				for time in execution_times.iter().rev() {
+					// Execution time is less than current time slot and in the past.  No more execution times need to be removed.
+					if *time < current_time_slot { 
+						break
+					}
+					// Execution time is equal to last time slot and task queue should be checked for task id.
+					// After checking task queue no other execution times need to be removed.
+					if *time == last_time_slot {
```
justinzhou93, line 834: why do we need to do this check with last_time_slot? if we just comb through the task_queue, it should be find the task_id if it's in there. and if not, then just need to comb through the different scheduled times i think?

also i worry about edge case during the start of the hour when the last_time_slot changes, if someone tries to remove then, before the first block builds and changes the last_time_slot to something new, if something weird happens there and we fail to remove from task queue

pallets/automation-time/src/lib.rs
```diff
@@ -787,52 +798,62 @@ pub mod pallet {
...
 		}
 
 		/// Removes the task of the provided task_id and all scheduled tasks, including those in the task queue.
 		///
-		fn remove_task(task_id: T::Hash, task: Task<T>) {
-			let current_time_slot = match Self::get_current_time_slot() {
-				Ok(time_slot) => time_slot,
-				Err(_) => return, // fix error
-			};
+		fn remove_task(task_id: T::Hash, task: Task<T>) -> Result<(), Error<T>> {
```
justinzhou93: oh huh, where did you add the new error? thought the errors in this function are bubbled up as events

pallets/automation-time/src/weights.rs
```diff
@@ -24,7 +24,7 @@
 // 100
 // --raw
 // --output
-// ./pallets/automation-time/src/weights.rs
+// ./pallets/automation-time/src/weights2.rs
```
justinzhou93, line 27: intentional?

pallets/automation-time/src/tests.rs
```diff
@@ -251,6 +251,35 @@ fn schedule_duplicates_errors() {
...
+}
+
+#[test]
+fn schedule_execution_times_removes_dupes() {
+	new_test_ext(START_BLOCK_TIME).execute_with(|| {
+		Balances::set_balance(RawOrigin::Root.into(), ALICE, 100_000, 5).unwrap();
+		assert_ok!(AutomationTime::schedule_notify_task(
+			Origin::signed(ALICE),
+			vec![50],
+			vec![SCHEDULED_TIME, SCHEDULED_TIME, SCHEDULED_TIME, SCHEDULED_TIME, SCHEDULED_TIME + 360],
```
justinzhou93, line 277: prob worth pulling the scheduled task to make sure only 2 times are in the execution times vector (and that they are what you expect)

pallets/automation-time/src/tests.rs
```diff
@@ -311,6 +340,70 @@ fn cancel_works_for_scheduled() {
...
+
+		assert_ok!(AutomationTime::cancel_task(Origin::signed(owner), task_id1,));
+
+		if let Some(_) = AutomationTime::get_scheduled_tasks(SCHEDULED_TIME) {
+			panic!("Tasks scheduled for the time it should have been deleted")
+		}
+		if let Some(_) = AutomationTime::get_scheduled_tasks(SCHEDULED_TIME + 120) {
+			panic!("Tasks scheduled for the time it should have been deleted")
+		}
+		if let Some(_) = AutomationTime::get_scheduled_tasks(SCHEDULED_TIME + 240) {
```
justinzhou93, line 392: prob also worth checking the tasks map to make sure that's not still in there as well

justinzhou93 (commented):
couple of things, but for the most part, looks really good!

oh and remember to add to changelog! like such: https://github.com/OAK-Foundation/OAK-blockchain/pull/127/commits/d710e1176d2fe4a81d77215adc8b4389798b41e8#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR8

pallets/automation-time/src/lib.rs
```diff
@@ -81,20 +81,31 @@ pub mod pallet {
...
 	#[scale_info(skip_type_params(T))]
 	pub struct Task<T: Config> {
 		owner_id: AccountOf<T>,
 		provided_id: Vec<u8>,
-		execution_times: BoundedVec<UnixTime, T::MaxRecurringTimes>,
+		execution_times: BoundedVec<UnixTime, T::MaxExecutionTimes>,
 		action: Action<T>,
 	}
 
+	impl<T: Config> PartialEq for Task<T> {
```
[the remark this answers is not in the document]
lreesby: yea, for business logic owner_id and provided_id make a unique task.  this was implemented so assert_eq would work and for the purpose of testing I thought we would want to know if the BoundedVec's were equal.

pallets/automation-time/src/lib.rs
lreesby: old commit?

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
justinzhou93: makes sense to me! thanks!

pallets/automation-time/src/lib.rs
```diff
@@ -765,33 +797,73 @@ pub mod pallet {
...
-							task_queue.remove(i);
-							TaskQueue::<T>::put(task_queue);
-							found_task = true;
-							break
-						}
+			let execution_times = Self::clean_execution_times_vector(task.execution_times.to_vec());
+			let current_time_slot = match Self::get_current_time_slot() {
+				Ok(time_slot) => time_slot,
+				// This will only occur for the first block in the chain.
+				Err(_) => 0,
```
lreesby, line 826: @rhuttman can you double check my thinking here?  if `get_current_time_slot` does not exist then we want to guarantee this block gets hit - https://github.com/OAK-Foundation/OAK-blockchain/pull/128/files#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R849-R866

rhuttman: that will work

pallets/automation-time/src/lib.rs
```diff
@@ -787,52 +798,62 @@ pub mod pallet {
...
-					Some(mut task_ids) =>
-						for i in 0..task_ids.len() {
-							if task_ids[i] == task_id {
-								if task_ids.len() == 1 {
-									<ScheduledTasks<T>>::remove(*time);
-								} else {
-									task_ids.remove(i);
-									<ScheduledTasks<T>>::insert(*time, task_ids);
-								}
+			if let Some((last_time_slot, _)) = Self::get_last_slot() {
```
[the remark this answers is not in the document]
lreesby: add `else` to loop through scheduled tasks.  https://github.com/OAK-Foundation/OAK-blockchain/pull/128/files#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R864-R881

pallets/automation-time/src/lib.rs
```diff
@@ -765,33 +796,73 @@ pub mod pallet {
 			<T as Config>::WeightInfo::run_native_transfer_task()
 		}
 
+		/// Determines if a task has been completed.
+		/// 
+		/// A task has been completed if all execution times exist in the past and/or are in the current time slot.
+		/// This means there a no more execution times to be run.
+		fn is_completed_task(execution_times: Vec<UnixTime>) -> bool {
```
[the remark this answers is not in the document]
lreesby: added check in run missing tasks.

pallets/automation-time/src/lib.rs
```diff
@@ -765,33 +796,73 @@ pub mod pallet {
...
+			if execution_times.len() == 0 { return true };
+
+			let current_time_slot = match Self::get_current_time_slot() {
+				Ok(time_slot) => time_slot,
+				// This will only occur for the first block in the chain.
+				Err(_) => return false,
+			};
+
+			let scheduled_times = Self::clean_execution_times_vector(execution_times);
```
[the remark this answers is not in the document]
lreesby: the dedup isn't too heavy as it just removes consecutive duplicates, which is why sort must happen first.

justinzhou93 (approved):
code in PR lgtm if you wanna merge and do separate PR for the missed queue stuff. but you can also just add that other missed queue stuff onto this PR as well

pallets/automation-time/src/lib.rs
```diff
@@ -796,22 +801,16 @@ pub mod pallet {
...
-
-			Self::clean_execution_times_vector(&mut execution_times);
-			let is_complete = *execution_times.last().unwrap() <= current_time_slot;
-			return is_complete;
+		/// Decrements task executions left.
+		/// If task is complete then removes task. If task not complete update task map.
+		/// A task has been completed if executions left equals 0.
+		fn decrement_task_and_remove_if_complete(task_id: T::Hash, mut task: Task<T>) {
+			task.executions_left = task.executions_left - 1;
+			if task.executions_left == 0 {
```
justinzhou93, line 809: less than or equal to. just in case haha

pallets/automation-time/src/lib.rs
```diff
@@ -796,22 +801,16 @@ pub mod pallet {
...
+				Tasks::<T>::remove(task_id);
+			} else {
+				Tasks::<T>::insert(task_id, task);
```
justinzhou93, line 812: prob test to make sure that this actually overwrites that task on the map. almost certain should, but check to make sure

pallets/automation-time/src/tests.rs
```diff
@@ -1057,7 +1187,7 @@ fn add_task_to_missed_queue(
 fn create_task(
 	owner: AccountId,
 	provided_id: Vec<u8>,
-	scheduled_time: u64,
+	scheduled_time: Vec<u64>,
```
justinzhou93, line 1190: nit update var name to plural

pallets/automation-time/src/lib.rs
```diff
@@ -711,15 +717,16 @@ pub mod pallet {
...
-						if Self::is_completed_task(task.execution_times.to_vec()) {
-							Tasks::<T>::remove(task_id);
-						}
 						task_action_weight
 							.saturating_add(T::DbWeight::get().writes(1 as Weight))
 							.saturating_add(T::DbWeight::get().reads(1 as Weight))
 					},
 				};
 
+				if let Some(task) = Self::get_task(task_id) {
```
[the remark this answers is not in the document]
justinzhou93: oh interesting. hmm, curious what's actually taking ownership of the task up there. i guess the action to construct NativeTransfer or Notify, it looks like it takes ownership of those fields.

i wonder if we could implement #[derive(Copy, Clone)] on the task struct (https://doc.rust-lang.org/rust-by-example/trait/clone.html), so that we can clone up there. alternatively, if you clone each of the fields that are going into the run_native_transfer_task or run_notify_task, i wonder if it'll let you keep ownership

justinzhou93 (commented):
couple of small things, but overall changes look good! the extra read on calling get_task again is unwieldy because it's incurring an extra read, which is about 25M weight, or 1/40000 of a second, which is rather heavy). 

although this is only doing the fix for incrementing index, right? this doesn't seem to include having the queue be filled with structs rather than task ids yet

justinzhou93 (commented):
1 nit, but otherwise looks good for the missed queue storing structs! but we prob still have to finish that ownership issue haha

pallets/automation-time/src/lib.rs
```diff
@@ -81,6 +81,20 @@ pub mod pallet {
 		NativeTransfer { sender: AccountOf<T>, recipient: AccountOf<T>, amount: BalanceOf<T> },
 	}
 
+	/// The enum that stores all action specific data.
+	#[derive(Debug, Eq, PartialEq, Encode, Decode, TypeInfo)]
+	#[scale_info(skip_type_params(T))]
+	pub struct MissedTask<T: Config> {
+		task_id: T::Hash,
+		execution_time: UnixTime,
```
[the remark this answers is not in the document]
rhuttman: we have to read the task anyway as we need to update/check `executions_left`.

pallets/automation-time/src/lib.rs
```diff
@@ -568,9 +616,15 @@ pub mod pallet {
 			last_time_slot: u64,
 		) -> (Weight, u64) {
 			if current_time_slot != last_time_slot {
-				let mut missed_tasks = Self::get_task_queue();
+				let missed_tasks = Self::get_task_queue();
 				let mut missed_queue = Self::get_missed_queue();
-				missed_queue.append(&mut missed_tasks);
+				for missed_task in missed_tasks {
```
rhuttman, line 621: We will need a way to do this without having to touch every item here by hand. We need thus to bulk moves

rhuttman (changes requested):

pallets/automation-time/src/lib.rs
```diff
@@ -633,7 +687,13 @@ pub mod pallet {
 			}
 			// Update the missed queue
 			let mut missed_queue = Self::get_missed_queue();
-			missed_queue.append(&mut missed_tasks);
+			for missed_task in missed_tasks {
+				let new_missed_task: MissedTask<T> = MissedTask::<T> {
+					task_id: missed_task,
+					execution_time: last_missed_slot,
```
rhuttman, line 693: this is not the correct time for all of the missed_tasks

justinzhou93: ahh crap, yah i missed this. but yah, this should be the right date: https://github.com/OAK-Foundation/OAK-blockchain/pull/128/files/20ee59cd998b39757533d248a51ff5605dc1508f#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R711

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
lreesby: also to get the `action` type to determine the weight.

pallets/automation-time/src/lib.rs
```diff
@@ -765,33 +827,83 @@ pub mod pallet {
 			<T as Config>::WeightInfo::run_native_transfer_task()
 		}
 
+		/// Decrements task executions left.
+		/// If task is complete then removes task. If task not complete update task map.
+		/// A task has been completed if executions left equals 0.
+		fn decrement_task_and_remove_if_complete(task_id: T::Hash, mut task: Task<T>) {
```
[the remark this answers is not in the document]
lreesby: `task_id` is not stored in the task object.  I could call `let task_id = Self::generate_task_id(owner_id.clone(), provided_id.clone());` to get the `task_id` and that would allow to not pass it in.
`

pallets/automation-time/src/tests.rs
```diff
@@ -309,12 +391,78 @@ fn cancel_works_for_scheduled() {
...
+		let task_id1 = schedule_task(
+			owner,
+			vec![40],
+			vec![SCHEDULED_TIME, SCHEDULED_TIME + 3600, SCHEDULED_TIME + 7200],
+			vec![2, 4, 5],
+		);
+		LastTimeSlot::<Test>::put((SCHEDULED_TIME - 14400, SCHEDULED_TIME - 14400));
+		System::reset_events();
+
+		match AutomationTime::get_scheduled_tasks(SCHEDULED_TIME) {
```
rhuttman, line 407: you test that this works earlier so don't need to check this again

pallets/automation-time/src/tests.rs
```diff
@@ -663,23 +898,94 @@ fn trigger_tasks_completes_some_native_transfer_tasks() {
...
+				assert_eq!(task.executions_left(), 2);
+			},
+		}
+
+		LastTimeSlot::<Test>::put((LAST_BLOCK_TIME, LAST_BLOCK_TIME));
+		System::reset_events();
+
+		AutomationTime::trigger_tasks(120_000);
+
+		match AutomationTime::get_task(task_id01) {
```
rhuttman, line 926: can you show that the task was run by checking the events as you did above. just makes it easier to read

pallets/automation-time/src/tests.rs
```diff
@@ -663,23 +898,94 @@ fn trigger_tasks_completes_some_native_transfer_tasks() {
...
+				assert_eq!(task.executions_left(), 1);
+			},
+		}
+
+		LastTimeSlot::<Test>::put((LAST_BLOCK_TIME, LAST_BLOCK_TIME));
+		System::reset_events();
+
+		AutomationTime::trigger_tasks(120_000);
+
+		assert_eq!(AutomationTime::get_task(task_id01), None);
```
rhuttman, line 962: can you show that the task was run by checking the events as you did above. just makes it easier to read

pallets/automation-time/src/tests.rs
```diff
@@ -663,23 +898,94 @@ fn trigger_tasks_completes_some_native_transfer_tasks() {
...
+			},
+			Some(task) => {
+				assert_eq!(task.executions_left(), 1);
+			},
+		}
+	})
+}
+
+#[test]
+fn trigger_tasks_removes_completed_tasks() {
```
rhuttman, line 938: also need this for missed

pallets/automation-time/src/tests.rs
```diff
@@ -309,12 +391,78 @@ fn cancel_works_for_scheduled() {
 	})
 }
 
+#[test]
+fn cancel_works_for_multiple_executions_scheduled() {
```
rhuttman, line 395: is there a test that checks for cancel when one of the task has already been executed?

pallets/automation-time/src/mock.rs
```diff
@@ -117,6 +117,8 @@ impl pallet_timestamp::Config for Test {
 
 parameter_types! {
 	pub const MaxTasksPerSlot: u32 = 2;
+	#[derive(Debug)]
```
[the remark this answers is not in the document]
lreesby: Need it for this:
````MaxExecutionTimes` doesn't implement `std::fmt::Debug`
the trait `std::fmt::Debug` is not implemented for `MaxExecutionTimes`
add `#[derive(Debug)]` to `MaxExecutionTimes` or manually `impl std::fmt::Debug for MaxExecutionTimes`

File: pallets/automation-time/src/weights.rs [hunk 35]
```diff
@@ -83,62 +85,76 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
...
-	// Storage: AutomationTime ScheduledTasks (r:1 w:0)
-	// Storage: AutomationTime TaskQueue (r:1 w:1)
-	fn force_cancel_overflow_task() -> Weight {
-		(23_000_000 as Weight)
-			.saturating_add(T::DbWeight::get().reads(3 as Weight))
-			.saturating_add(T::DbWeight::get().writes(2 as Weight))
+	// Storage: Timestamp Now (r:1 w:0)
+	// Storage: AutomationTime LastTimeSlot (r:1 w:0)
+	// Storage: AutomationTime ScheduledTasks (r:24 w:0)
+	fn force_cancel_overflow_task_full() -> Weight {
```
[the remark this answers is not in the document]
lreesby: because this test never scheduled the task just put one straight on the task queue.  so it is looking for all executions times and not finding any.  https://github.com/OAK-Foundation/OAK-blockchain/pull/128/files/2d244949e795dd25c452201e06a24a89e1ed9474#diff-eb0580064fc58a5d272e21ab88621358160d3d67ea4a20f6fea4161373e5e15fR198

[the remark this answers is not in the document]
lreesby: fixed.

pallets/automation-time/src/weights.rs
```diff
@@ -83,62 +85,76 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
...
 	// Storage: AutomationTime Tasks (r:1 w:1)
+	// Storage: Timestamp Now (r:1 w:0)
+	// Storage: AutomationTime LastTimeSlot (r:1 w:0)
 	// Storage: AutomationTime ScheduledTasks (r:1 w:1)
 	fn cancel_scheduled_task() -> Weight {
-		(14_000_000 as Weight)
-			.saturating_add(T::DbWeight::get().reads(2 as Weight))
+		(18_000_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(4 as Weight))
 			.saturating_add(T::DbWeight::get().writes(2 as Weight))
```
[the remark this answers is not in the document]
lreesby: This benchmark doesn't get used anywhere but as the benchmark is written it's testing a task with a single execution time.  An execution time can only be in either the trigger map or the task queue, not both.  Removing the benchmark, though, since it's not being used.

pallets/automation-time/src/mock.rs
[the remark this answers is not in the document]
rhuttman: `{:?}` uses `std::fmt::Debug`  to know how to print

rhuttman (approved):

justinzhou93 (approved):
lgtm. ~make sure @rhuttman is cool with all the changes as well~ ignore, i'm late

rhuttman (approved):
lgtm

PR #129, opened 2022-04-15
NEU xcm config in separate file

rhuttman:


justinzhou93 (approved):
we need to do the same for turing? or just separate PR later?

rhuttman:
I am expecting to do a lot more stuff to NEU to get it to where I want. I will update TUR when I have all the pieces

PR #131, opened 2022-04-21
Use parachain-staking for collator selection, delegation, and inflation-based payouts

nvengal:
This pr replaces collator-selection with parachain-staking. This migrates our chain from PoA (proof of authority) to dPoS (delegated proof of stake). Parachain-staking handles the mechanics of collator bonding and delegation, selection of the active set of collators, and payouts (through inflation) for successful block production.

### Basics

Parachain Staking sees the world through `rounds`. A `round` is a set of blocks with the same active set of collators. Currently 25 blocks/round or 5 minutes for Neumann.

Each round a new active set is determined by selecting n collators with the highest total stake (including delegations).
image

At the beginning of each round collator and delegator rewards are paid out with 1 collator and associated delegators being paid out each block until everyone is paid. From a given payout 20% is reserved for the collator, 30% for the parachain bond fund, and the remaining 50% is split across stakers proportional to their stake. There is a configurable delay (# of rounds) before rewards are paid out.
image

### How to become a collator candidate (numbers based on current proposal):

- Call `parachainStaking_joinCandidates` with at least the MinCollatorBond (500 NEU) to enter the candidatePool
- In order to become an eligible collator you'll need to hit the MinCollatorStk (1 MNEU). You can do this either by increasing your bond (candidateBondMore) or acquiring delegators.
- Delegators can bond to a candidate using `parachainStaking_delegate` with atleast 50 NEU
image
image

### Notes on inflation/rewards

- Inflation defaults to 0. This allows us to keep rewards off until we're ready to enable them. Inflation is enabled using sudo with `parachainStaking_setStakingExpectations` and `parachainStaking_setInflation`. Note: Inflation is set using parts per billion.
- Until the parachain bond fund account is set through sudo `parachainStaking_setParachainBondAccount` the percentage payout that would have been reserved for the parachain bond goes to stakers.

### Session pallet, cumulus, nimbus

To reduce the size of this change and avoid a hard fork we opted to stick with cumulus concensus rather than migrate to nimbus. We've had to fork the parachain-staking pallet from moonbeam to implement the traits required by the session and authorship pallets.

Our fork of parachain-staking begins a new session every time the round is changed so we can think of 1 session == 1 round. On each new session the new set of active collators is passed to the session pallet to further filter specific authors for a given slot. We also implemented `note_author` from the authorship pallet to track and award collators for the blocks which they've authored.

__Important Note__: Since the session pallet is being used in conjunction with parachain-staking potential collators will need to register with both pallets to successfully author blocks. The session pallet requires registering the session key from `author_rotateKeys` with `session_setKeys`. If none of the active collator set from parachain-staking has registered keys with the session pallet the chain will brick.

### Runtime Upgrade

After the runtime upgrade the parachain-staking candidate pool will be empty. As long as the candidate pool is empty parachain-staking will be unable to pass any new collators to the session pallet so the session pallet will continue to use its existing set of collators. We will need to transfer funds to our current invulnerable collators so that they can bond and participate in the pos system.

### DIFF
--- a/node/src/chain_spec/neumann.rs
+++ b/node/src/chain_spec/neumann.rs
@@ -376,8 +376,6 @@ fn testnet_genesis(
 		},
 		balances: neumann_runtime::BalancesConfig { balances: endowed_accounts },
 		parachain_info: neumann_runtime::ParachainInfoConfig { parachain_id: para_id },
-		// Defaults to active collators from session pallet unless configured otherwise
-		parachain_staking: Default::default(),
 		session: neumann_runtime::SessionConfig {
 			keys: invulnerables
 				.into_iter()
@@ -390,6 +388,8 @@ fn testnet_genesis(
 				})
 				.collect(),
 		},
+		// Defaults to active collators from session pallet unless configured otherwise
+		parachain_staking: Default::default(),
 		// no need to pass anything to aura, in fact it will panic if we do. Session will take care
 		// of this.
 		aura: Default::default(),
--- a/node/src/chain_spec/neumann.rs
+++ b/node/src/chain_spec/neumann.rs
@@ -376,9 +376,9 @@ pub fn neumann_inflation_config() -> InflationInfo<Balance> {
 	InflationInfo {
 		// staking expectations
 		expect: Range {
-			min: 100_000 * 1_000_000_000_000_000_000,
-			ideal: 200_000 * 1_000_000_000_000_000_000,
-			max: 500_000 * 1_000_000_000_000_000_000,
+			min: 4_000_000 * DOLLAR,
+			ideal: 6_000_000 * DOLLAR,
+			max: 8_000_000 * DOLLAR,
 		},
 		annual,
 		round: to_round_inflation(annual),
--- a/node/src/chain_spec/neumann.rs
+++ b/node/src/chain_spec/neumann.rs
@@ -11,8 +11,8 @@ use crate::chain_spec::{
 	Extensions,
 };
 use neumann_runtime::{
-	CouncilConfig, MinCollatorStk, SudoConfig, TechnicalMembershipConfig, ValveConfig,
-	VestingConfig, DOLLAR, EXISTENTIAL_DEPOSIT, TOKEN_DECIMALS,
+	CouncilConfig, SudoConfig, TechnicalMembershipConfig, ValveConfig, VestingConfig, DOLLAR,
+	EXISTENTIAL_DEPOSIT, TOKEN_DECIMALS,
 };
 use parachain_staking::{inflation, InflationInfo, Range};
 use primitives::{AccountId, AuraId, Balance};
@@ -61,8 +61,6 @@ pub fn development_config() -> ChainSpec {
 			let endowed_accounts: Vec<(AccountId, Balance)> =
 				accounts.iter().cloned().map(|k| (k, initial_balance)).collect();
 
-			let collator_bond = MinCollatorStk::get();
-
 			testnet_genesis(
 				// initial collators.
 				vec![
@@ -82,7 +80,6 @@ pub fn development_config() -> ChainSpec {
 				vec![],
 				vec![get_account_id_from_seed::<sr25519::Public>("Alice")],
 				vec![get_account_id_from_seed::<sr25519::Public>("Alice")],
-				collator_bond,
 			)
 		},
 		Vec::new(),
@@ -129,8 +126,6 @@ pub fn local_testnet_config() -> ChainSpec {
 			let initial_vesting: Vec<(u64, Vec<(AccountId, Balance)>)> =
 				serde_json::from_slice(vesting_json).unwrap();
 
-			let collator_bond = MinCollatorStk::get();
-
 			testnet_genesis(
 				// initial collators.
 				vec![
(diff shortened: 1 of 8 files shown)


runtime/neumann/src/lib.rs
```diff
@@ -971,12 +990,14 @@ impl_runtime_apis! {
 			use pallet_automation_time::Pallet as AutomationTime;
 			use pallet_valve::Pallet as Valve;
 			use pallet_vesting::Pallet as Vesting;
+			use parachain_staking::Pallet as ParachainStaking;
```
nvengal, line 993: Very unclear on how to use the runtime benchmarks. Only included this here because I saw other projects doing the same.

nvengal:
Been testing this manually a lot but looking for feedback on what other tests might be useful (and where to add them). Decently concerned about the upgrade process and still testing it out but happy to take any feedback or direction.

runtime/neumann/src/lib.rs
```diff
@@ -110,6 +110,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
+	parachain_staking::InitGenesisOnRuntimeUpgrade<Runtime>,
```
nvengal, line 113: This migrations attempts to replicate some of the storage defaults that would have been set if this pallet had been present at genesis.

nvengal:
CI will fail till https://github.com/OAK-Foundation/moonbeam/pull/2 is merged

runtime/neumann/src/lib.rs
```diff
@@ -775,7 +794,7 @@ impl Contains<Call> for ClosedCallFilter {
 			Call::AutomationTime(_) => false,
 			Call::Balances(_) => false,
 			Call::Bounties(_) => false,
-			Call::CollatorSelection(_) => false,
+			Call::ParachainStaking(_) => false,
```
irsal, line 797: if we close the parachain staking gate, what happens?

irsal (commented):
* 5 min rounds make sense to me
* 1 session = 1 round; this means that we're shortening a given session from 1200 blocks down to 25 blocks? is that accurate?

Any other NEU configs worth flagging here?

node/src/chain_spec/neumann.rs
```diff
@@ -389,11 +376,8 @@ fn testnet_genesis(
 		},
 		balances: neumann_runtime::BalancesConfig { balances: endowed_accounts },
 		parachain_info: neumann_runtime::ParachainInfoConfig { parachain_id: para_id },
-		collator_selection: neumann_runtime::CollatorSelectionConfig {
-			invulnerables: invulnerables.iter().cloned().map(|(acc, _)| acc).collect(),
-			candidacy_bond: collator_bond,
-			..Default::default()
-		},
+		// Defaults to active collators from session pallet unless configured otherwise
+		parachain_staking: Default::default(),
```
rhuttman, line 380: won't this only be used when standing up new NEU test chains now? Shouldn't we configure this to something that makes sense for that?

runtime/neumann/src/lib.rs
rhuttman: when do we need to remove this? when we do the next upgrade?

justinzhou93: yah from what i've seen, once we're sure this is in all of the chains correctly, other chains remove their migration

runtime/neumann/src/lib.rs
nvengal: I believe that this just shuts off access to the parachain_staking extrinsics for non sudo accounts. @rhuttman?

node/src/chain_spec/neumann.rs
nvengal: We can definitely configure this to something that makes sense for that but I'm not sure what that is. Leaving it as `Default::default()` means that parachain staking is going to defer to the session pallet/ invulnerables. Regardless of whether this stays as default or not I'd prefer to make that change in a separate pr since it won't have any bearing on the runtime upgrade.

runtime/neumann/src/lib.rs
nvengal: Ideally we remove it before the next upgrade. Even if we don't this migration won't cause any harm by being run more than once.

nvengal:
@irsal 

> this means that we're shortening a given session from 1200 blocks down to 25 blocks? is that accurate?

yup, you've understood it correctly. I'm not aware of any issues with having short sessions, but maybe we want it to be a bit longer than 25 blocks.

> Any other NEU configs worth flagging here?

The only other particularly notable config is the `1 million neu` minimum collator stake. I'm not sure about what the distribution of neumann looks like, but I imagine we want this number to be sufficiently high that our existing invulnerable collators are still a part of each authoring set (for the short term).

runtime/neumann/src/lib.rs
nvengal: Removed this in favor of using the on runtime hook in the pallet

justinzhou93:
wait, don't our current collators only have 400k tokens right now? if we bump to 1M, does that kick out our current collators?

node/src/chain_spec/neumann.rs
nvengal: Made this issue to track this work- https://oak-network.atlassian.net/browse/ENG-360

runtime/neumann/src/lib.rs
rhuttman: correct

rhuttman (approved):

nvengal:
Added turing config

rhuttman (approved):

PR #135, opened 2022-05-03
fix benchmarks for greater than 256

justinzhou93:


runtime/neumann/src/lib.rs
```diff
@@ -727,7 +727,7 @@ impl pallet_democracy::Config for Runtime {
 }
 
 parameter_types! {
-	pub const MaxTasksPerSlot: u32 = 256;
+	pub const MaxTasksPerSlot: u32 = 576;
```
justinzhou93, line 730: don't have to bump spec_version in this PR, right? we'll just do in the next release

pallets/automation-time/src/weights.rs
```diff
@@ -106,7 +106,7 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
 	// Storage: AutomationTime LastTimeSlot (r:1 w:0)
 	// Storage: AutomationTime ScheduledTasks (r:24 w:24)
 	fn cancel_scheduled_task_full() -> Weight {
-		(542_000_000 as Weight)
+		(1_192_000_000 as Weight)
```
justinzhou93, line 109: lordie this got heavy

pallets/automation-time/src/weights.rs
```diff
@@ -124,7 +124,7 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
 	// Storage: AutomationTime LastTimeSlot (r:1 w:0)
 	// Storage: AutomationTime ScheduledTasks (r:24 w:24)
 	fn force_cancel_scheduled_task_full() -> Weight {
-		(524_000_000 as Weight)
+		(1_203_000_000 as Weight)
```
justinzhou93, line 127: this too. damn, cancels got heavy

justinzhou93:
key here is turning an array of a single u8 number into an array of 2 u8 numbers. since (u8, u8) should offer the same number of probabilities as u16 (2^8 * 2^8 = 2^16)

lreesby (approved):
lgtm

PR #136, opened 2022-05-09
moving turing xcm configs to their own file

rhuttman:


runtime/turing/src/xcm_config.rs
```diff
@@ -0,0 +1,178 @@
+use super::{
```
justinzhou93, line 1: hmm do we need to import this specifically into the runtime file? or is it being imported through this line: https://github.com/OAK-Foundation/OAK-blockchain/pull/136/commits/9e4afab563788f9cfe9702fe9ee6344ca52802ad#diff-5a1af4b45c1636993d2a09dc36f37fd0be817e6aea96166a690753c2c707d801R67?

justinzhou93 (approved):

runtime/turing/src/xcm_config.rs
rhuttman: it is imported by that yes

PR #139, opened 2022-05-11
0.19 

rhuttman:


node/src/command.rs
```diff
@@ -309,21 +310,45 @@ pub fn run() -> Result<()> {
...
-					{
-						runner.sync_run(|config| cmd.run::<Block, Executor>(config))
+		Some(Subcommand::Benchmark(cmd)) => {
+			let runner = cli.create_runner(cmd)?;
+			let chain_spec = &runner.config().chain_spec;
+			// Switch on the concrete benchmark sub-command-
+			with_runtime_or_err!(chain_spec, {
+				{
+					match cmd {
+						BenchmarkCmd::Pallet(cmd) =>
```
rhuttman, line 320: This is realllllyyyy important to note as it changes the benchmark command. You now need to add `pallet` as a subcommand:

./target/release/oak-collator benchmark pallet \
--chain dev \
--execution wasm \
--wasm-execution compiled \
--pallet pallet_valve \
--extrinsic "*" \
--repeat 64 \
--steps 20 \
--output ./pallets/valve/src/raw-weights.rs

node/src/command.rs
```diff
@@ -309,21 +310,45 @@ pub fn run() -> Result<()> {
...
+							let partials = service::new_partial::<RuntimeApi, Executor, _>(
+								&config,
+								crate::service::parachain_build_import_queue,
+							)?;
+							let db = partials.backend.expose_db();
+							let storage = partials.backend.expose_storage();
+
+							cmd.run(config, partials.client.clone(), db, storage)
+						}),
+						BenchmarkCmd::Overhead(_) => Err("Unsupported benchmarking command".into()),
```
justinzhou93, line 347: what is block, storage and overhead? (well, overhead does nothing, but what about the other two?)

node/src/command.rs
justinzhou93: might be good to run benchmarks so the command is saved in the weights file? or just add it manually, or in a wiki or something

pallets/vesting/src/benchmarking.rs
```diff
@@ -14,8 +14,6 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-#![cfg(feature = "runtime-benchmarks")]
```
justinzhou93, line 17: how come we don't need anymore? does it default to not compiling now?

runtime/neumann/src/xcm_config.rs
```diff
@@ -85,19 +86,85 @@ parameter_types! {
...
-pub type Barrier = (
-	TakeWeightCredit,
-	AllowTopLevelPaidExecutionFrom<Everything>,
-	AllowUnpaidExecutionFrom<ParentOrParentsExecutivePlurality>,
-	// ^^^ Parent and its exec plurality get free execution
-);
+//TODO: move DenyThenTry to polkadot's xcm module.
+/// Deny executing the xcm message if it matches any of the Deny filter regardless of anything else.
+/// If it passes the Deny, and matches one of the Allow cases then it is let through.
+pub struct DenyThenTry<Deny, Allow>(PhantomData<Deny>, PhantomData<Allow>)
```
justinzhou93, line 99: very slight chance i'll have to bug you for when i merge this into your branch for xcm examples

justinzhou93 (approved):

pallets/vesting/src/benchmarking.rs
rhuttman: I moved it to the lib file to be consistent w/ the test and mock. I need to do a general cleanup of all the files like this but I was in here messing around trying to get something to compile so it caught my eye

node/src/command.rs
rhuttman: I am not sure honestly. The documentation has not been updated yet

runtime/neumann/src/xcm_config.rs
rhuttman: I will be testing this heavily in my PR for the cross token stuff right after this merges. Just nothing to test it with really rn

lreesby (approved):
lgtm

PR #140, opened 2022-05-12
Allow storing of other tokens and cross chain token transfers

rhuttman:


runtime/neumann/Cargo.toml
```diff
@@ -91,6 +91,14 @@ xcm-executor = { git = "https://github.com/paritytech/polkadot", default-feature
 # Moonbeam Dependencies
 parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", default-features = false, branch = "oak-polkadot-v0.9.19" }
 
+# ORML Dependencies
```
irsal, line 94: another version dependency to keep track of 🙃

rhuttman:
Tested locally:
- Transferring Kusd, KAR, LKSM, and TUR between chains

runtime/neumann/src/xcm_config.rs
```diff
@@ -166,19 +173,70 @@ pub type Barrier = DenyThenTry<
...
+        if let MultiAsset {
+            id: AssetId::Concrete(id),
+            fun: Fungibility::Fungible(amount),
+        } = revenue
+        {
+            if let Some(currency_id) = CurrencyIdConvert::convert(id) {
+                // Ensure TreasuryAccount have ed requirement for native asset, but don't need
+				// ed requirement for cross-chain asset because it's one of whitelist accounts.
+				// Ignore the result.
+				let _ = Currencies::deposit(currency_id, &TreasuryAccount::get(), amount);
```
nvengal, line 219: nit: formatting

runtime/turing/src/xcm_config.rs
```diff
@@ -165,19 +172,106 @@ pub type Barrier = DenyThenTry<
...
+        if let MultiAsset {
+            id: AssetId::Concrete(id),
+            fun: Fungibility::Fungible(amount),
+        } = revenue
+        {
+            if let Some(currency_id) = CurrencyIdConvert::convert(id) {
+                // Ensure TreasuryAccount have ed requirement for native asset, but don't need
+				// ed requirement for cross-chain asset because it's one of whitelist accounts.
+				// Ignore the result.
+				let _ = Currencies::deposit(currency_id, &TreasuryAccount::get(), amount);
```
nvengal, line 247: nit: formatting.

Is it worth logging or something if this fails here? This seems like the sort of thing that we wouldn't just want to fail silently.

nvengal (approved):

runtime/neumann/src/xcm_config.rs
rhuttman: believe it or not the format looks correct in my IDE, so not sure what is going on here

runtime/turing/src/xcm_config.rs
rhuttman: I need to think about how to handle these sorts of things. Will add later this week

PR #145, opened 2022-05-17
Bump runtime spec version to 281

nvengal:
Adds CollatorRegistration to parachain staking config

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [1.3.0] - 2022-05-17
 - Moving Turing xcm configs to their own file.
 - Allow foreign tokens to get stored on-chain
 - Enable cross chain token transfers
 - Lower existential deposit from .1 to .01
 - Adding Karura tokens
+- Ensure CollatorRegistration before joining candidate pool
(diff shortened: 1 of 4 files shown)


lreesby (approved):
lgtm

runtime/neumann/src/lib.rs
```diff
@@ -143,7 +143,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("neumann"),
 	impl_name: create_runtime_str!("neumann"),
 	authoring_version: 1,
-	spec_version: 280,
+	spec_version: 281,
```
nvengal, line 146: Not sure if there are other versions that need to bumped alongside this

justinzhou93 (approved):
remember to add to the changelog release notes! should be in a section called "Unreleased". also for future, i wonder if it's better to use git tags rather than branches, since branches can desync, whereas tags don't change

runtime/neumann/src/lib.rs
irsal: `transaction_version` i think

justinzhou93 (approved):
lgtm

CHANGELOG.md
```diff
@@ -4,12 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
```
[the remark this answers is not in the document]
nvengal: can do

justinzhou93 (approved):
🙏  thanks for the changes to all the nit picks haha

PR #146, opened 2022-05-17
Schedule an xcmp task

lreesby:


node/src/chain_spec/neumann.rs
```diff
@@ -307,6 +310,7 @@ fn testnet_genesis(
 			phantom: Default::default(),
 		},
 		parachain_system: Default::default(),
+		polkadot_xcm: PolkadotXcmConfig { safe_xcm_version: Some(SAFE_XCM_VERSION) },
```
justinzhou93, line 313: i'm guessing this is just for us to decide on whether we're using version 0-2, right? i'm curious what it's actually used for in polkadot. the only place where i could find it being used was https://github.com/paritytech/polkadot/blob/master/xcm/pallet-xcm/src/lib.rs#L1224 and i have no idea what it's doing there

pallets/automation-time/src/tests.rs
```diff
@@ -231,19 +236,92 @@ fn schedule_native_transfer_works() {
...
+		get_funds(Sibling::from(PARA_ID).into_account());
+		assert_ok!(AutomationTime::schedule_xcmp_task(
+			cumulus_pallet_xcm::Origin::SiblingParachain(PARA_ID.try_into().unwrap()).into(),
+			vec![50],
+			vec![SCHEDULED_TIME],
+			PARA_ID.try_into().unwrap(),
+			call.clone(),
+			100_000,
+		));
+		match AutomationTime::get_scheduled_tasks(SCHEDULED_TIME) {
```
justinzhou93, line 252: wonder if it's worthwhile checking the tasks map as well, to make sure it's correctly typed as type "XCMP" rather than normal notify or something else

pallets/automation-time/src/tests.rs
```diff
@@ -231,19 +236,92 @@ fn schedule_native_transfer_works() {
...
+	})
+}
+
+#[test]
+fn schedule_xcmp_errors_para_id_mismatch() {
+	new_test_ext(START_BLOCK_TIME).execute_with(|| {
+		let para_id_2: u32 = 2001;
+		assert_noop!(
+			AutomationTime::schedule_xcmp_task(
+				cumulus_pallet_xcm::Origin::SiblingParachain(PARA_ID.try_into().unwrap()).into(),
```
justinzhou93, line 283: ahh i vaguely remember you having some t rouble with origin type conversion. What was the other origin type? was it XCM OriginType or something from Substrate? Either way, I'm curious how you got past the type conversion?

pallets/automation-time/src/tests.rs
```diff
@@ -231,19 +236,92 @@ fn schedule_native_transfer_works() {
...
+				vec![3, 4, 5],
+				100_000,
+			),
+			Error::<Test>::ParaIdMismatch,
+		);
+	})
+}
+
+#[test]
+fn schedule_xcmp_errors_bad_origin() {
```
justinzhou93, line 296: should we have a variety of cases here for Bad Origin? i.e. i think there are a bunch of ways to trigger it, right?
1. if the sovereign account does not match the Para ID
2. If the OriginType is SovereignAccount rather than Native
3. May be some other way to trigger that I don't know?

pallets/automation-time/src/tests.rs
```diff
@@ -1017,30 +1124,101 @@ fn missed_tasks_removes_completed_tasks() {
...
+		System::reset_events();
+
+		AutomationTime::trigger_tasks(120_000);
+
+		assert_eq!(
+			sent_xcm(),
+			vec![(
+				(1, Junction::Parachain(PARA_ID.into())).into(),
+				Xcm(vec![Transact {
+					origin_type: OriginKind::SovereignAccount,
```
justinzhou93, line 1178: wait, shouldn't this origin_type be type Native? I had trouble getting SovereignAccount to work at least while doing testing locally. This could also totally be a test only thing, but would be good to know why i think?

pallets/automation-time/src/tests.rs
```diff
@@ -1017,30 +1124,101 @@ fn missed_tasks_removes_completed_tasks() {
...
+				para_id: PARA_ID.try_into().unwrap(),
+				task_id: task_id1
+			}),]
+		);
+	})
+}
+
+#[test]
+fn trigger_tasks_xcmp_sends_error_event() {
+	new_test_ext(START_BLOCK_TIME).execute_with(|| {
```
justinzhou93, line 1196: might be good to add a comment or note here reminding us about what's causing this to fail. Is it because it's a faulty call that's being stored in XCMP action?

pallets/automation-time/src/tests.rs
```diff
@@ -1340,7 +1518,7 @@ fn schedule_task(
 }
 
 fn add_task_to_task_queue(
-	owner: AccountId,
+	owner: u8,
```
justinzhou93, line 1521: hm curious why owner is a u8. do you mean to say [u8, 32]?

pallets/automation-time/src/weights.rs
```diff
@@ -100,6 +102,19 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
...
 	}
+	// Storage: Timestamp Now (r:1 w:0)
+	// Storage: System Account (r:2 w:2)
+	// Storage: AutomationTime Tasks (r:1 w:1)
+	// Storage: AutomationTime ScheduledTasks (r:1 w:1)
+	fn schedule_xcmp_task_full(v: u32, ) -> Weight {
+		(30_331_000 as Weight)
+			// Standard Error: 99_000
+			.saturating_add((49_354_000 as Weight).saturating_mul(v as Weight))
+			.saturating_add(T::DbWeight::get().reads(4 as Weight))
```
justinzhou93, line 113: ah i forget. the extra 1 read 1 write from "unknown" isn't in this PR yet, right? if you did manage to fix it tho, i'm curious how. but i think last we chatted, you just hadn't pushed that piece up yet?

pallets/automation-time/src/tests.rs
lreesby: 1. this is tested here `schedule_xcmp_errors_para_id_mismatch` and returns the error `ParaIdMismatch`
2. I fixed this test and am passing a `SovereignAccount` and that returns `BadOrigin`

pallets/automation-time/src/tests.rs
lreesby: yea, this should be a `SovereignAccount`.  I was playing around and broke it.  Fixed now.

pallets/automation-time/src/tests.rs
lreesby: I commented in the mock file but can add a comments here, too.  https://github.com/OAK-Foundation/OAK-blockchain/blob/laura/schedule-xcmp-task/pallets/automation-time/src/mock.rs#L149

pallets/automation-time/src/tests.rs
lreesby: `cumulus_pallet_xcm::Origin::SiblingParachain(PARA_ID.try_into().unwrap()).into()`  === `Native`

you can also do it another way by setting the inner and outer call.
```
let dispatch_origin = XcmOriginToCallOrigin::convert_origin(
	MultiLocation { parents: 1, interior: X1(Parachain(PARA_ID)) },
	OriginKind::SovereignAccount,
).unwrap();
let call: OuterCall = Call::schedule_xcmp_task { 
	provided_id: vec![50],
	execution_times: vec![SCHEDULED_TIME + 1],
	para_id: para_id2.try_into().unwrap(),
	call: vec![4,5,6],
}.into();
assert_ok!(call.dispatch(dispatch_origin));

File: pallets/automation-time/src/weights.rs (hunk 8 above)
Comment (lreesby) [reply pallets/automation-time/src/weights.rs]: needed to push this.

File: pallets/automation-time/src/tests.rs (hunk 2 above)
Comment (lreesby) [reply pallets/automation-time/src/tests.rs:252]: This test does that.  It gets all the scheduled taskId's, which should only be 1, then it grabs the task from the task map and compares it to what we expect it to be.

File: node/src/chain_spec/neumann.rs (hunk 1 above)
Comment (lreesby) [reply node/src/chain_spec/neumann.rs:313]: that function checks the receiving network for the supported version and then returns the minimum version number between the sending and receiving networks.  pretty much enforcing backwards compatibility but not forwards.  from the this article (https://polkadot.network/blog/xcm-part-two-versioning-and-compatibility/):
"It must therefore be the responsibility of the sending network to ensure that the message it sends is capable of being interpreted by the receiving network. In concrete terms, the version of XCM used for the message must be no newer than the version of XCM that the receiving network supports."

File: pallets/automation-time/src/tests.rs (hunk 2 above)
Comment (justinzhou93) [reply pallets/automation-time/src/tests.rs:252]: gah, if i just read that more closely. i misread the next 2 lines 🤦

File: pallets/automation-time/src/tests.rs (hunk 3 above)
Comment (justinzhou93) [reply pallets/automation-time/src/tests.rs:283]: ahhh i see i see. thanks!!

**Review by justinzhou93 — APPROVED** [review]
don't forget to merge master! but lgtm otherwise

File: pallets/automation-time/src/lib.rs [hunk 9]
```diff
@@ -312,6 +348,10 @@ pub mod pallet {
 			task_id: T::Hash,
 			execution_time: UnixTime,
 		},
+
+		XCMPTest {
```
rhuttman, line 352: is this still needed?

atomaka:
@lreesby I added a no-op commit to trigger the pipeline.

justinzhou93 (approved):

pallets/automation-time/src/lib.rs
```diff
@@ -949,32 +1059,23 @@ pub mod pallet {
...
-									vec![task_id].try_into().unwrap();
-								<ScheduledTasks<T>>::insert(*time, task_ids);
-							},
-							Some(mut task_ids) => {
-								if let Err(_) = task_ids.try_push(task_id) {
-									return Rollback(Err(DispatchError::Other("time slot full")))
-								}
-								<ScheduledTasks<T>>::insert(*time, task_ids);
-							},
+			for time in execution_times.iter() {
```
nvengal, line 1062: Trying to understand this diff to resolve a merge conflict.

Isn't the `with_transaction` pretty relevant here so that we don't partially schedule a user's task while returning an error?

lreesby: This line makes it a transaction. 

https://github.com/OAK-Foundation/OAK-blockchain/pull/146/files#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R399

PR #151, opened 2022-05-21
Adding Parallel Heiko tokens to Turing

rhuttman:


node/src/chain_spec/turing.rs
```diff
@@ -74,7 +73,7 @@ pub fn turing_development_config() -> ChainSpec {
 				],
 				get_account_id_from_seed::<sr25519::Public>("Alice"),
 				endowed_accounts,
-				DEFAULT_PARA_ID.into(),
+				REGISTERED_PARA_ID.into(),
```
nvengal, line 76: Any particular reason for this change? It ends up being one more thing you have to keep in your head when running the turing-dev chain locally. At face value it seems easier to just leave the default for all dev/local chains at 2000.

nvengal: Oh, does this have side effects for testing xcm locally in some way that I'm not aware of?

rhuttman: exactly. This could be the reason why a dev and local chain are needed I guess. When making XCM calls to other chains they will have us coded in as paraId 2114. This means we either need to change our paraId or change their code.

runtime/turing/src/lib.rs
```diff
@@ -343,6 +343,8 @@ parameter_type_with_key! {
 			CurrencyId::KUSD => CurrencyId::KUSD.cent(),
 			CurrencyId::KAR => 10 * CurrencyId::KAR.cent(),
 			CurrencyId::LKSM => 50 * CurrencyId::LKSM.millicent(),
+			CurrencyId::HKO => 50 * CurrencyId::HKO.cent(),
```
lreesby, line 346: double checking this should be `50`?  the others all seem to be `10`.

lreesby (approved):
lgtm

nvengal (approved):

justinzhou93 (approved):

justinzhou93 (approved):

PR #159, opened 2022-05-25
Use unvested funds as additional issuance for parachain staking inflation calcs

nvengal:
Added a storage item to the vesting pallet to keep track of total amount of unvested funds.
Vesting also implements the trait required by parachain-staking to return any additional allocation for calculating inflation.

Tested on neumann-local chain with 1 BNEU base issuance:
- Inflation no vesting: 1,860,000,000,019 (per round)
- Inflation with 1 BNEU vesting: 3,720,000,000,019 (per round)

### DIFF
--- a/node/src/chain_spec/neumann.rs
+++ b/node/src/chain_spec/neumann.rs
@@ -121,12 +121,12 @@ pub fn local_testnet_config() -> ChainSpec {
 				accounts.iter().cloned().map(|k| (k, initial_balance)).collect();
 
 			let now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_secs();
-			let now_to_the_minute = now - (now % 3600);
+			let now_to_the_hour = now - (now % 3600);
 			let month_in_seconds = 2_628_000;
 			let vest_amount = 1_000_000_000_000;
 			let initial_vesting: Vec<(u64, Vec<(AccountId, Balance)>)> = vec![
 				(
-					now_to_the_minute + month_in_seconds,
+					now_to_the_hour + month_in_seconds,
 					vec![
 						(accounts[0].clone(), vest_amount),
 						(accounts[1].clone(), vest_amount),
@@ -134,7 +134,7 @@ pub fn local_testnet_config() -> ChainSpec {
 					],
 				),
 				(
-					now_to_the_minute + 2 * month_in_seconds,
+					now_to_the_hour + 2 * month_in_seconds,
 					vec![
 						(accounts[0].clone(), vest_amount),
 						(accounts[1].clone(), vest_amount),


pallets/vesting/src/lib.rs
```diff
@@ -157,8 +183,10 @@ pub mod pallet {
 						"Cannot vest less than the existential deposit"
 					);
 					scheduled_vests.push((account.clone(), amount.clone()));
+					unvested_issuance = unvested_issuance.saturating_add(*amount);
 				}
 				VestingSchedule::<T>::insert(time, scheduled_vests);
+				TotalUnvestedIssuance::<T>::set(unvested_issuance);
```
nvengal, line 189: Noticed that the api has both `put` and `set` but it's unclear to me what the difference is.

pallets/vesting/src/weights.rs
```diff
@@ -2,12 +2,13 @@
 //! Autogenerated weights for `pallet_vesting`
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2022-03-01, STEPS: `20`, REPEAT: 64, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2022-05-25, STEPS: `20`, REPEAT: 64, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
nvengal, line 5: No idea if I did this right. Just ran the command in comments and hoped for the best. The number of reads/writes look reasonable but I'm not sure how to make sense of the const values.

pallets/vesting/src/lib.rs
```diff
@@ -91,12 +92,20 @@ pub mod pallet {
...
+
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(_: T::BlockNumber) -> Weight {
 			let vest_count = Self::vest();
 			<T as Config>::WeightInfo::vest(vest_count)
 		}
+
+		fn on_runtime_upgrade() -> Weight {
+			migrations::set_total_unvested_issuance::<T>()
```
nvengal, line 107: This is idempotent so I didn't bother implementing storage version checks.

node/Cargo.toml
```diff
@@ -111,4 +111,4 @@ polkadot-service = { git = "https://github.com/paritytech/polkadot", branch = "r
 polkadot-test-service = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20" }
 
 # Moonbeam Dependencies
-parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", default-features = false, branch = "oak-polkadot-v0.9.20" }
+parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", default-features = false, branch = "nikhil/unvested-issuance" }
```
nvengal, line 114: TODO: point parachain-staking at release branch once https://github.com/OAK-Foundation/moonbeam/pull/9 is merged

node/src/chain_spec/neumann.rs
```diff
@@ -119,10 +120,28 @@ pub fn local_testnet_config() -> ChainSpec {
 			let endowed_accounts: Vec<(AccountId, Balance)> =
 				accounts.iter().cloned().map(|k| (k, initial_balance)).collect();
 
-			let vesting_json =
-				&include_bytes!("../../../distribution/neumann_vesting_local.json")[..];
```
nvengal, line 123: Opted to do vesting config in rust so that timestamps can be dynamic and in the future. The timestamps in this file are from march. I wasn't sure who owned the accounts in the json file so opted to use the dev seeds (alice, bob, etc) for vesting too.

nvengal:
This is ready for review but shouldn't be merged until the dependent parachain-staking pr is merged and the cargo toml links are moved back to the release branch.

pallets/vesting/src/lib.rs
```diff
@@ -91,12 +94,20 @@ pub mod pallet {
...
+
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(_: T::BlockNumber) -> Weight {
 			let vest_count = Self::vest();
 			<T as Config>::WeightInfo::vest(vest_count)
 		}
+
+		fn on_runtime_upgrade() -> Weight {
+			migrations::set_total_unvested_allocation::<T>()
```
justinzhou93, line 109: do you have to gate with a storage version and/or update the storage version for next time? e.g. https://github.com/OAK-Foundation/OAK-blockchain/blob/master/pallets/automation-time/src/lib.rs#L69

we don't have to, but we'd just have to be pretty careful about removing afterwards

pallets/vesting/src/migrations.rs
```diff
@@ -0,0 +1,21 @@
...
+use log::info;
+use sp_runtime::traits::{Saturating, Zero};
+
+pub fn set_total_unvested_allocation<T: Config>() -> Weight {
+	info!(target: "vesting", "Setting TotalUnvestedAllocation");
+
+	let mut reads = 0;
+	let unvested_funds = VestingSchedule::<T>::iter()
+		.map(|(_, vests)| {
+			reads = reads.saturating_add(1);
```
justinzhou93, line 12: oh hm, they don't just remove the whole thing from the DB in 1 read? somehow i always thought they bulk removed and then processed all in memory, although i guess they'd run outta memory pretty quick that way. hm makes sense.

justinzhou93 (approved):
code lgtm, just 1 quick thing about storage versions

pallets/vesting/src/migrations.rs
nvengal: I'm not totally sure that this is the case. I based my code off of
> Furthermore, because accessing the elements of a map requires more database reads than accessing the elements of a native list, map iterations are significantly more costly than list iterations in terms of execution time.
(https://docs.substrate.io/v3/runtime/storage/)

but it's not totally clear that this is what they meant.

rhuttman (approved):

PR #162, opened 2022-05-26
Use self-hosted runner for test suite

atomaka:


imstar15:
Excuse me, should we keep the test-coverage?

atomaka:
> Excuse me, should we keep the test-coverage?

Yup! We should. I was just wanted to get a basic test build passing so that I could benchmark against some self-hosted nodes I'm going to test.

.github/workflows/check.yml
```diff
@@ -1,57 +1,39 @@
...
-      - name: Test coverage
-        run: |
-          sh ./.maintain/test-coverage/test-coverage.sh
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Setup Rust
+      run: rustup show
+    - name: Build
+      run: |
+        cargo build --release --all
```
nvengal, line 36:
```suggestion
        cargo build --release --locked --all-features
```
We may want to consider further limiting this to exclude the runtime-benchmarks features and anything else that might be low value but time consuming. But this is a good start regardless.

.github/workflows/check.yml
```diff
@@ -1,57 +1,39 @@
...
+    - name: Test
+      run: |
+        cargo test --release --all
```
nvengal, line 39:
```suggestion
        cargo test --release --locked --all-features
```

rust-toolchain.toml
```diff
@@ -1,3 +1,5 @@
 [toolchain]
 channel = "nightly-2021-12-15"
+components = ["rustfmt"]
 targets = ["wasm32-unknown-unknown"]
+profile = "minimal"
```
nvengal, line 5: I don't fully know the implications of this. Looks like minimal is the right choice for CI but setting it here will also affect our local development.

Could explicitly set the profile for CI in the action
```
rustup set profile minimal
```

nvengal: Was referencing this doc- https://rust-lang.github.io/rustup/concepts/profiles.html

PR #164, opened 2022-05-31
Increase turing minimum candidate self-bond

nvengal:
Increasing the candidate self bond from 400,000 to 2,000,000 tur in prep for enabling community collators outside the closed sale.

### DIFF
--- a/node/src/chain_spec/turing.rs
+++ b/node/src/chain_spec/turing.rs
@@ -105,6 +105,10 @@ fn testnet_genesis(
 	general_councils: Vec<AccountId>,
 	technical_memberships: Vec<AccountId>,
 ) -> turing_runtime::GenesisConfig {
+	let candidate_stake = std::cmp::max(
+		turing_runtime::MinCollatorStk::get(),
+		turing_runtime::MinCandidateStk::get(),
+	);
 	turing_runtime::GenesisConfig {
 		system: turing_runtime::SystemConfig {
 			code: turing_runtime::WASM_BINARY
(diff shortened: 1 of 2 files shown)


node/src/chain_spec/turing.rs
```diff
@@ -105,6 +105,10 @@ fn testnet_genesis(
 	general_councils: Vec<AccountId>,
 	technical_memberships: Vec<AccountId>,
 ) -> turing_runtime::GenesisConfig {
+	let candidate_stake = std::cmp::max(
```
justinzhou93, line 108: why do max for the two? shouldn't MinCandidateStk always be larger? Or is it because of some weird edgecase in the runtime where MinCandidateStk could not yet be set and therefore be 0?

justinzhou93 (approved):
1 q, otherwise lgtm!

node/src/chain_spec/turing.rs
nvengal: Generally, MinCandidateStk is lower that MinCollatorStk. The CandidateStk is the amount that the collator has to put up themselves, while the CollatorStk is the minimum required to be selected to author blocks. A Collator's Stk includes their CandidateStk + any delegations so generally it should be a larger number. The relationship is thrown off a bit for us because we've allowed some collators to join the candidate pool at a reduced number in the closed sale and we want to make sure that they can still produce blocks once this goes live (so MinCollatorStk has to stay at the lower number.

nvengal:
@irsal Heads up, this is the change to bump the min self-bond to 2 million tur. Any closed sale collators that plan to join at 400,000 will need to get in before this is live.

irsal (approved):
thx!

PR #173, opened 2022-06-14
Correct formatter errors

atomaka:
Allow `cargo fmt -- --check` to pass.

`cargo fmt` created build errors in `node/src/command.rs` so two lines now have formatting skipped.


rhuttman (approved):

rhuttman (changes requested):
Need to fix PR check issue with "#![rustfmt::skip]"

atomaka:
> Need to fix PR check issue with "#![rustfmt::skip]"

Ugh. Due to old nightly in our `rust-toolchain.toml` file. We should decide what versions of Rust we want to use at some point and update the configuration accordingly. BUT for now, I can just cheat the linter and invert the two features.

nvengal (approved):

PR #174, opened 2022-06-14
update benchmarks automation pallet

lreesby:


justinzhou93 (approved):
oh hm curious how come deposit multiplier is so high. is it cause ryan bumped the execution fee weights? but otherwise lgtm

lreesby:
> oh hm curious how come deposit multiplier is so high. is it cause ryan bumped the execution fee weights? but otherwise lgtm

correct.  I made it super high to keep it from failing in the future.  with the asserts we should at least know if it does.

rhuttman (approved):

PR #176, opened 2022-06-15
xcm changes for OAK

justinzhou93:


node/src/chain_spec/turing.rs
```diff
@@ -72,7 +72,7 @@ pub fn turing_development_config() -> ChainSpec {
 				get_account_id_from_seed::<sr25519::Public>("Alice"),
 				endowed_accounts,
 				REGISTERED_PARA_ID.into(),
-				vec![b"AutomationTime".to_vec(), b"Balances".to_vec(), b"Democracy".to_vec()],
```
justinzhou93, line 75: removed these for dev cause it sucks to have to make a call each time to get them to just open up

justinzhou93:
just FYI, this is built off of laura's PR - so that prob has to merge first

pallets/automation-time/src/lib.rs
```diff
@@ -344,6 +344,9 @@ pub mod pallet {
 			task_id: T::Hash,
 			execution_time: UnixTime,
 		},
+		XCMPTest {
+			account_id: AccountId,
+		},
```
lreesby, line 349: This should be removed.

runtime/turing/src/lib.rs
```diff
@@ -402,6 +403,7 @@ parameter_type_with_key! {
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub enum CurrencyId {
 	Native,
+	UNIT,
```
rhuttman, line 406: you can;t change the order. It's append only for the enum

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Turing staging chain spec added
```
rhuttman, line 8: I don't think these are supposed to be here

runtime/turing/src/lib.rs
```diff
@@ -373,6 +373,7 @@ parameter_type_with_key! {
 	pub ExistentialDeposits: |currency_id: CurrencyId| -> Balance {
 		match currency_id {
 			CurrencyId::Native => EXISTENTIAL_DEPOSIT,
+			CurrencyId::UNIT => 10 * CurrencyId::UNIT.millicent(),
```
rhuttman, line 376: Do we need this token for Turing?

rhuttman (changes requested):

runtime/turing/src/lib.rs
justinzhou93: hm i recall we were deprecating neumann and only keeping it around so that it's easier for us to restructure our code with multiple chain feature flags once we get OAK. so i've mainly been doing development on turing-dev. so with that, i imagined that it would probably be good to have this currency be there as well. would that be fine or should be just keep it contained to neumann for now?

rhuttman (approved):

PR #179, opened 2022-06-17
Build Github workflow for release

atomaka:
Add manually release workflow that can be triggered via the Github actions interface.

Screen Shot 2022-06-16 at 21 13 54

See https://github.com/OAK-Foundation/ci-blockchain/actions/workflows/release.yml for example runs:
* https://github.com/OAK-Foundation/ci-blockchain/actions/runs/2506172789 with Runtime WASM build
* https://github.com/OAK-Foundation/ci-blockchain/actions/runs/2536846279 with Debian Package

This will:
* Build the Rust binary using our self-hosted runner
* Build a Debian package
* Build the WASM runtime using srtool in a Github hosted runner
* Build the Docker image by copying the built Rust binary and upload it to Docker hub
* Generate a draft release with appropriate notes and attach built artifacts

This does not create the git tag and requires it to manually generated before release.

To consider:
* Release note message format
* Including checksum for oak-collator binary
* Improving Docker image

Example output from draft:

Screen Shot 2022-06-22 at 18 45 18

Example run: https://github.com/OAK-Foundation/OAK-blockchain/actions/runs/2545381877


.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+      - name: Download binary
+        uses: actions/download-artifact@v2
+        with:
+          name: oak-collator
+      - name: Build
+        run: |
+          docker build -f ./docker/turing/Dockerfile -t oaknetwork/oak-testchain:${{ github.event.inputs.tag }} .
+          docker tag oaknetwork/oak-testchain:${{ github.event.inputs.tag }} oaknetwork/oak-testchain:latest
+          docker push oaknetwork/oak-testchain:${{ github.event.inputs.tag }}
+          docker push oaknetwork/oak-testchain:latest
```
nvengal, line 82: Assuming that `oak-testchain` is a leftover value from testing this in another repo.

atomaka: Yup. We'll update before merge, but I don't want to ship Docker images to a repository I don't have access to delete in.

docker/turing/Dockerfile
```diff
@@ -30,16 +18,15 @@ RUN apt-get update && apt-get install -y curl && \
   mkdir /data && \
   chown -R $USERNAME:$USERNAME /data && \
   ln -s /data /$USERNAME/.local/share/$BINARY && \
-  find /usr/bin/* /usr/sbin/* | grep -v curl | xargs rm
+  find /usr/bin/* /usr/sbin/* | grep -v 'curl\|sh\|chmod' | xargs rm
```
atomaka, line 21: This pattern includes a lot more than we'd want (`sh` matches many things).  I think our Docker image should be reviewed a bit more:

1. Smaller base image (not Alpine because of musl libc; maybe Debian), but maybe Debian's spent effort in making a minimal image since I've last used
2. Second image maybe? Ideally, we'd only need to run a `COPY` and a `RUN chmod` in CI.
3. Consider running our Debian pacakge.

I don't think we need to solve those here though..but we could.

node/Cargo.toml
```diff
@@ -112,3 +112,15 @@ polkadot-test-service = { git = "https://github.com/paritytech/polkadot", branch
...
 # Moonbeam Dependencies
 parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", default-features = false, branch = "oak-polkadot-v0.9.20" }
+
+[package.metadata.deb]
+name = "oak-collator"
+maintainer-scripts = "../.maintain/debian/deb-maintainer-scripts"
+assets = [
+	["target/release/oak-collator", "/usr/bin/", "755"],
+	["../.maintain/debian/oak-collator.service", "/lib/systemd/system/", "644"],
+	["../.maintain/debian/defaults", "/usr/share/oak-collator/oak-collator", "644"]
```
atomaka, line 122: No idea why these relative paths are different; verify this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,160 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.tag }}
+      - name: Download binary
+        uses: actions/download-artifact@v2
+        with:
+          name: oak-collator
```
nvengal, line 81: Isn't this a zip with both the binary and the deb file? Doesn't it need to be unpackaged before doing the docker build?
I didn't see anything in the download-artifact action documentation about this implicitly unpacking but maybe it does.

node/Cargo.toml
nvengal: That is strange. To verify this do I need to build and run this on a debian box?

node/Cargo.toml
```diff
@@ -112,3 +112,15 @@ polkadot-test-service = { git = "https://github.com/paritytech/polkadot", branch
 
 # Moonbeam Dependencies
 parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", default-features = false, branch = "oak-polkadot-v0.9.20" }
+
+[package.metadata.deb]
```
nvengal, line 116: Any value in adding any of the systemd specific config units?

node/Cargo.toml
atomaka: To verify, you can:

1. Grab the package I generated in the test repository
    * https://github.com/OAK-Foundation/ci-blockchain/releases/tag/untagged-a68ac2ab076874d6c05d
2. Inspect it on a linux box
    * `mkdir -p /tmp/oak; dpkg -x oak-collator.deb /tmp/oak`
3. Install the package and verify binary, user/group, service, and configuration file is present
    * `sudo dpkg -i oak-collator.deb`
4. Make sure the service starts
    * `sudo systemctl start oak-collator.service`

I've run through #4, but plan to sanity check myself when I build the test release from this repository.

.github/workflows/release.yml
atomaka: The download-artifact action is doing this for us.

node/Cargo.toml
atomaka: I didn't see these configuration options so...there might be?

nvengal (approved):

PR #180, opened 2022-06-17
Add automation time auto compound stake action

nvengal:
This PR adds a new automation time action, `AutoCompoundDelegatedStake`, that calls out parachain-staking to increase a delegators stake up to a fixed minimum. Notably, this task reschedules itself before completion using a user-provided frequency.

On error, the task emits an event and should not reschedule. Some errors are recoverable (the selected scheduled_tasks bucket is full) while others are not (no longer delegating). For now, we're treating all errors as non-recoverable but in the future we should consider strategies that will keep the task running in the case of recoverable errors.

- [x] Fix Cargo.tomls after https://github.com/OAK-Foundation/moonbeam/pull/14 is merged (CI will not pass until this is done)


pallets/automation-time/src/benchmarking.rs
```diff
@@ -173,6 +210,17 @@ benchmarks! {
 		let provided_id: Vec<u8> = vec![(T::MaxTasksPerSlot::get()/256).try_into().unwrap(), (T::MaxTasksPerSlot::get()%256).try_into().unwrap()];
 	}: schedule_native_transfer_task(RawOrigin::Signed(caller), provided_id, times, recipient, transfer_amount)
 
+	schedule_auto_compound_delegated_stake_task_full {
```
nvengal, line 213: My first time doing benchmarking so worth a good look here to make sure I did it right.

pallets/automation-time/src/lib.rs
```diff
@@ -955,6 +1089,80 @@ pub mod pallet {
...
+			let _ = Self::reschedule_existing_task(
+				task_id,
+				task.owner_id.clone(),
+				&task.action,
+				new_execution_times.clone(),
+			)
+			.map(|_| {
+				let new_executions_left: u32 = new_execution_times.len().try_into().unwrap();
+				task.executions_left += new_executions_left;
+				new_execution_times.iter().try_for_each(|t| task.execution_times.try_push(*t))
```
nvengal, line 1136: It didn't seem like the execution times that are attached to a task actually do anything, but it seemed appropriate to keep them in sync anyway.

pallets/automation-time/src/lib.rs
```diff
@@ -955,6 +1089,80 @@ pub mod pallet {
...
+					return (
+						task,
+						<T as Config>::WeightInfo::run_auto_compound_delegated_stake_task(),
+					)
+				},
+			}
+
+			let new_execution_times: Vec<UnixTime> =
+				task.execution_times.iter().map(|when| when.saturating_add(frequency)).collect();
+			let _ = Self::reschedule_existing_task(
```
nvengal, line 1127: I'm not thrilled about my implementation here as it requires updating the task afterwards to ensure that all the metadata like `execution_times` and `executions_left` stays correct. In the long (hopefully short) run, I'm hoping that all of these methods that act on the task data can be moved to the task impl so that state management becomes easier. Anyway, I'm not thrilled with my implementation and very open to improvements.

pallets/automation-time/src/lib.rs
```diff
@@ -955,6 +1089,80 @@ pub mod pallet {
...
+					}),
+				Err(e) => {
+					Self::deposit_event(Event::AutoCompoundDelegatorStakeFailed {
+						task_id,
+						error_message: Into::<&str>::into(e).as_bytes().to_vec(),
+						error: e,
+					});
+					return (
+						task,
+						<T as Config>::WeightInfo::run_auto_compound_delegated_stake_task(),
```
nvengal, line 1120: I could technically return a smaller weight here, I think. Is that something that I should be doing?

pallets/automation-time/src/lib.rs
```diff
@@ -1065,23 +1272,36 @@ pub mod pallet {
...
+		}
+
+		/// Insert task id into scheduled tasks
+		/// With transaction will protect against a partial success where N of M execution times might be full,
+		/// rolling back any successful insertions into the schedule task table.
+		fn insert_scheduled_tasks(
+			task_id: T::Hash,
+			execution_times: Vec<UnixTime>,
+		) -> Result<T::Hash, Error<T>> {
+			with_transaction(|| -> storage::TransactionOutcome<Result<T::Hash, DispatchError>> {
```
nvengal, line 1285: @lreesby flagging this for your review in particular since you removed `with_transaction` in one of your pr. While trying to resolve the merge conflict I ended up putting it back in, because I'm calling this same method outside of the extrinsic context where I only want to revert a small part of the transaction.

pallets/automation-time/src/lib.rs
```diff
@@ -1129,12 +1350,43 @@ pub mod pallet {
...
 			T::Hashing::hash_of(&task_hash_input)
 		}
 
+		pub fn generate_auto_compound_delegated_stake_provided_id(
+			delegator: &AccountOf<T>,
+			collator: &AccountOf<T>,
+		) -> Vec<u8> {
+			let mut provided_id = "AutoCompoundDelegatedStake".as_bytes().to_vec();
+			provided_id.extend(delegator.encode());
+			provided_id.extend(collator.encode());
```
nvengal, line 1386: I had a lot of trouble getting the account ids to convert into something that I could use for a provided id and ended up here. It's not clear to me what the `encode` method is doing other than that it converts the account id to a byte string. I expected the resulting hex to be the public key of the account but it doesn't seem to be. Let me know if any of you know any less opaque conversions.

pallets/automation-time/src/tests.rs
```diff
@@ -309,6 +309,67 @@ fn schedule_xcmp_errors_bad_origin() {
...
+				AccountId32::new(BOB),
+				100_000,
+			),
+			Error::<Test>::InvalidTime,
+		);
+	})
+}
+
+#[test]
+fn schedule_auto_compound_with_high_frequency() {
```
justinzhou93, line 358: might not need, since we test for this in the is_valid_time, which should already have been tested earlier

pallets/automation-time/src/tests.rs
```diff
@@ -309,6 +309,67 @@ fn schedule_xcmp_errors_bad_origin() {
...
+				3_600,
+				bob,
+				1_000_000_000,
+			))
+		);
+	})
+}
+
+#[test]
+fn schedule_auto_compound_with_bad_frequency() {
```
justinzhou93, line 342: might not need, since we test for this in the is_valid_time, which should already have been tested earlier

justinzhou93 (commented):
mostly nitpicks

pallets/automation-time/src/lib.rs
```diff
@@ -963,6 +1171,7 @@ pub mod pallet {
 			if task.executions_left <= 0 {
 				Tasks::<T>::remove(task_id);
 			} else {
+				task.execution_times.remove(0);
```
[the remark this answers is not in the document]
rhuttman: The goal here was to keep the execution times for posterity. We need to think about if that is really needed or not.

pallets/automation-time/src/lib.rs
```diff
@@ -955,6 +1090,80 @@ pub mod pallet {
...
+			});
+
+			(task, <T as Config>::WeightInfo::run_auto_compound_delegated_stake_task())
+		}
+
+		fn compound_delegator_stake(
+			delegator: T::AccountId,
+			collator: T::AccountId,
+			account_minimum: BalanceOf<T>,
+			execution_fee: Result<BalanceOf<T>, DispatchError>,
```
rhuttman, line 1155: seems kind of odd to handle this potential error like this

pallets/automation-time/src/lib.rs
```diff
@@ -955,6 +1090,80 @@ pub mod pallet {
...
+		) -> Result<BalanceOf<T>, DispatchErrorWithPostInfo> {
+			let reserved_funds = account_minimum.saturating_add(execution_fee?);
```
rhuttman, line 1157: they already paid for this execution, so why would we not up their bond then not schedule the next?

rhuttman (commented):
I would like to talk about the larger decision to lump auto-reschedule in with the action of upping a bond on a delegator

node/Cargo.toml
```diff
@@ -113,7 +113,9 @@ polkadot-test-service = { git = "https://github.com/paritytech/polkadot", branch
 xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.23" } 
 
 # Moonbeam Dependencies
-pallet-parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", branch = "oak-polkadot-v0.9.23" }
+# TODO: fix dep
+pallet-parachain-staking = { path = "/Users/nikhil/oak/moonbeam/pallets/parachain-staking" }
+# pallet-parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", branch = "oak-polkadot-v0.9.23" }
```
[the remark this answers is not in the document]
nvengal: yup, will switch it to the right location once that pr is merged.

pallets/automation-time/src/lib.rs
```diff
@@ -298,6 +359,8 @@ pub mod pallet {
 		TooManyExecutionsTimes,
 		/// ParaId provided does not match origin paraId.
 		ParaIdMismatch,
+		/// Account minimum not met
+		AccountMinimumBalanceNotMet,
```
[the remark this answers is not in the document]
nvengal: Good catch. Those do seem duplicative, I'll reuse the InsufficientBalance.

pallets/automation-time/src/lib.rs
```diff
@@ -844,6 +960,24 @@ pub mod pallet {
...
+								frequency,
+							} => {
+								let (mut_task, weight) =
+									Self::run_auto_compound_delegated_stake_task(
+										delegator,
+										collator,
+										account_minimum,
+										frequency,
+										task_id.clone(),
+										task,
```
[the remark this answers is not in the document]
nvengal: yeah, this is clunky. it's a good thought that maybe there's a different data structure that we could be interacting with here.

pallets/automation-time/src/tests.rs
nvengal: the way I've written it, `is_valid_time` is by being bypassed for the `frequency`

pallets/automation-time/src/lib.rs
lreesby: this line would be redundant then and can be removed for each extrinsic call.  https://github.com/OAK-Foundation/OAK-blockchain/pull/180/files#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R601

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
nvengal: I went ahead and moved the `task.execution_times.remove(0)` in the new task that I added to preserve the current behavior of the task struct.

pallets/automation-time/src/lib.rs
nvengal: Yeah! Fully agreed. But rather than writing a bunch of code to handle that what I think needs to happen is that this function should go back to using `saturating_*` instead of throwing errors: https://github.com/OAK-Foundation/OAK-blockchain/blob/d7b7824f918acda4093c5faad7c3154cba8897b2/pallets/automation-time/src/lib.rs#L1153-L1160

I'm not sure why we changed this to checked+error here. It leads to overly verbose or unnecessary downstream code as you noticed. This method calculates an execution fee. It is reasonable for us to saturate the execution fee on `BalanceOf`. Balance is the same type that's used for total_issuance. We're never overflowing it and even if we did, we should just default to the biggest number.

I'm happy to change the `calculate_execution_fee` method in another pr if my reasoning seems sound.

pallets/automation-time/src/lib.rs
nvengal: See- https://github.com/OAK-Foundation/OAK-blockchain/pull/180#discussion_r923901660

This method will never throw an error.

nvengal:
Addressed PR feedback

justinzhou93 (approved):
lgtm, thanks for the changes!

pallets/automation-time/src/lib.rs
nvengal: My mistake on what you meant, added todo for this edge- https://github.com/OAK-Foundation/OAK-blockchain/pull/180/commits/8c00dc65f1a0624959ff976cf951be6198ef010f

rhuttman (approved):

pallets/automation-time/src/lib.rs
```diff
@@ -508,6 +576,49 @@ pub mod pallet {
...
 		}
 
+		/// Schedule a task to increase delegation to a specified up to a minimum balance
+		/// Task will reschedule itself to run on a given frequency until a failure occurs
+		///
+		/// # Parameters
+		/// * `execution_time`: The unix timestamp when the task should run for the first time
+		/// * `frequency`: Number of seconds to wait inbetween task executions
+		/// * `collator_id`: Account ID of the target collator
+		/// * `account_minimum`: The minimum amount of funds that should be left in the wallet
```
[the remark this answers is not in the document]
nvengal: Good concern, but I think we're safe since these funds aren't getting spent, just reserved. So the account doesn't get reaped. I couldn't find documentation to back me up, but I managed to test it locally-
image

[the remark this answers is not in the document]
nvengal: Also verified that rewards still get paid out when you have 0 transferrable balance- 
image

PR #187, opened 2022-06-29
Fix saturating calculation inaccuracy.

imstar15:
image

Working on it, not finished yet.

### DIFF
--- a/pallets/automation-time/src/lib.rs
+++ b/pallets/automation-time/src/lib.rs
@@ -57,8 +57,8 @@ use log::info;
 use pallet_timestamp::{self as timestamp};
 use scale_info::TypeInfo;
 use sp_runtime::{
-	traits::{SaturatedConversion, Saturating},
-	DispatchError, Perbill,
+	traits::{SaturatedConversion, Saturating, CheckedMul},
+	DispatchError, Perbill, ArithmeticError,
 };
 use sp_std::{vec, vec::Vec};
 
@@ -984,7 +984,7 @@ pub mod pallet {
 			who: T::AccountId,
 			provided_id: Vec<u8>,
 			mut execution_times: Vec<UnixTime>,
-		) -> Result<(), Error<T>> {
+		) -> Result<(), DispatchError> {
 			if provided_id.len() == 0 {
 				Err(Error::<T>::EmptyProvidedId)?
 			}
@@ -998,9 +998,9 @@ pub mod pallet {
 			}
 
 			let fee =
-				Self::calculate_execution_fee(&action, execution_times.len().try_into().unwrap());
+				Self::calculate_execution_fee(&action, execution_times.len().try_into().unwrap())?;
 			T::NativeTokenExchange::can_pay_fee(&who, fee.clone())
-				.map_err(|_| Error::InsufficientBalance)?;
+				.map_err(|_| Error::<T>::InsufficientBalance)?;
 
 			let task_id =
 				Self::schedule_task(who.clone(), provided_id.clone(), execution_times.clone())?;
@@ -1016,7 +1016,7 @@ pub mod pallet {
 
 			// This should never error if can_pay_fee passed.
 			T::NativeTokenExchange::withdraw_fee(&who, fee.clone())
-				.map_err(|_| Error::LiquidityRestrictions)?;
+				.map_err(|_| Error::<T>::LiquidityRestrictions)?;
 
 			Self::deposit_event(Event::TaskScheduled { who, task_id });
 			Ok(())
@@ -1028,16 +1028,16 @@ pub mod pallet {
 			T::Hashing::hash_of(&task_hash_input)
 		}
 
-		fn calculate_execution_fee(action: &Action<T>, executions: u32) -> BalanceOf<T> {
+		fn calculate_execution_fee(action: &Action<T>, executions: u32) -> Result<BalanceOf<T>, DispatchError> {
 			let action_weight = match action {
 				Action::Notify { message: _ } => <T as Config>::WeightInfo::run_notify_task(),
 				Action::NativeTransfer { sender: _, recipient: _, amount: _ } =>
 					<T as Config>::WeightInfo::run_native_transfer_task(),
 			};
 
-			let total_weight = action_weight.saturating_mul(executions.into());
+			let total_weight = action_weight.checked_mul(executions.into()).ok_or(ArithmeticError::Overflow)?;
 			let weight_as_balance = <BalanceOf<T>>::saturated_from(total_weight);
-			T::ExecutionWeightFee::get().saturating_mul(weight_as_balance)
+			Ok(T::ExecutionWeightFee::get().checked_mul(&weight_as_balance).ok_or(ArithmeticError::Overflow)?)
 		}
 	}
 }


justinzhou93 (approved):

imstar15:
I think a distinction needs to be made between two cases:

1. If it is an extrinsic function or a function on its call chain, then we use the checked_* function to do the calculation and throw a DispatchError.

2. If it is the on_initialize hook and its function on the call chain, we cannot throw Error. Because it is a defined callback function. 

```
fn on_initialize(_: T::BlockNumber) -> Weight {
	...
}
```

Therefore, we can only continue to use the calculation function of saturating_*.

Here I refer to the code of substrate.

https://github.com/paritytech/substrate/blob/f39769740858bf480ef13adbfd46267afce92eb9/frame/contracts/src/lib.rs#L402

```
fn on_initialize(_block: T::BlockNumber) -> Weight {
...
		let weight_limit = T::BlockWeights::get()
			.max_block
			.saturating_sub(System::<T>::block_weight().total())
			.min(T::DeletionWeightLimit::get());
...
}
```

pallets/automation-time/src/lib.rs
```diff
@@ -562,14 +562,14 @@ pub mod pallet {
...
+		pub fn get_current_time_slot() -> Result<UnixTime, DispatchError> {
 			let now = <timestamp::Pallet<T>>::get().saturated_into::<UnixTime>();
 			if now == 0 {
 				Err(Error::<T>::BlockTimeNotSet)?
 			}
-			let now = now / 1000;
-			let diff_to_hour = now % 3600;
-			Ok(now.saturating_sub(diff_to_hour))
+			let now = now.checked_div(1000).ok_or(ArithmeticError::Overflow)?;
+			let diff_to_hour = now.checked_rem(3600).ok_or(ArithmeticError::Overflow)?;
```
justinzhou93, line 578: doesn't checked_* functions return Option<number>? do you have to unwrap that option?

pallets/automation-time/src/lib.rs
```diff
@@ -562,14 +565,18 @@ pub mod pallet {
 		/// * Get the most recent timestamp from the block.
 		/// * Convert the ms unix timestamp to seconds.
 		/// * Bring the timestamp down to the last whole hour.
-		pub fn get_current_time_slot() -> Result<UnixTime, Error<T>> {
-			let now = <timestamp::Pallet<T>>::get().saturated_into::<UnixTime>();
+		pub fn get_current_time_slot() -> Result<UnixTime, DispatchError> {
```
justinzhou93, line 568: is this dispatch error correct? can't it also be Error<T> as well? does this not currently break compile for line 574?

pallets/automation-time/src/lib.rs
```diff
@@ -578,13 +585,13 @@ pub mod pallet {
 		/// - End in a whole hour
 		/// - Be in the future
 		/// - Not be more than MaxScheduleSeconds out
-		fn is_valid_time(scheduled_time: UnixTime) -> Result<(), Error<T>> {
+		fn is_valid_time(scheduled_time: UnixTime) -> Result<(), DispatchError> {
```
justinzhou93, line 588: same here as above for dispatch error

pallets/automation-time/src/tests_calculation.rs
```diff
@@ -0,0 +1,212 @@
+// This file is part of OAK Blockchain.
```
justinzhou93, line 1: curious why redo the runtime for tests here. does the existing mock file for automation time not work?

justinzhou93 (changes requested):

pallets/automation-time/src/lib.rs
imstar15: Doesn't checked_* functions return Option? 
checked_div return Option: `Some()` or `None`.

do you have to unwrap that option?
Yes, I have to unwrap the option.

When it is None, dispatch ArithmeticError to tell the user that an error was encountered during processing.
Here is the code in substrate, which I referenced.

https://github.com/paritytech/substrate/blob/6a946fc36d68b89599d7ca1ab03803d10c78468c/frame/assets/src/functions.rs#L71

https://github.com/paritytech/substrate/blob/6a946fc36d68b89599d7ca1ab03803d10c78468c/frame/balances/src/lib.rs#L1033

pallets/automation-time/src/lib.rs
imstar15: is this dispatch error correct?
Yes, it's OK!

can't it also be Error as well? 
NO, If it is Error here, the compilation will fail.
Error<T> does not contains ArithmeticError.

https://github.com/OAK-Foundation/OAK-blockchain/blob/6b5e7c9158028539079f959e4b3f9bae1b68d917/pallets/automation-time/src/lib.rs#L268

Refers to substrate code:
https://github.com/paritytech/substrate/blob/6a946fc36d68b89599d7ca1ab03803d10c78468c/frame/balances/src/lib.rs#L1001

does this not currently break compile for line 574?
No, It doesn't break compile.

pallets/automation-time/src/lib.rs
imstar15: https://github.com/OAK-Foundation/OAK-blockchain/pull/187#discussion_r915399384

pallets/automation-time/src/lib.rs
```diff
@@ -1085,7 +1096,7 @@ pub mod pallet {
 			who: T::AccountId,
 			provided_id: Vec<u8>,
 			mut execution_times: Vec<UnixTime>,
-		) -> Result<(), Error<T>> {
+		) -> Result<(), DispatchError> {
```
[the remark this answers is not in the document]
imstar15: https://github.com/OAK-Foundation/OAK-blockchain/pull/187#discussion_r915399384

pallets/automation-time/src/tests_calculation.rs
imstar15: I want to change the value of the constant `MaxScheduleSeconds`.

https://github.com/OAK-Foundation/OAK-blockchain/blob/6b5e7c9158028539079f959e4b3f9bae1b68d917/pallets/automation-time/src/tests_calculation.rs#L129

Otherwise, the `ArithmeticError::Overflow` case cannot be tested.

Is there a better way? please give me some advices.

pallets/automation-time/src/lib.rs
justinzhou93: ahh you're good! my bad, i missed the question mark

pallets/automation-time/src/lib.rs
justinzhou93: hm interesting. i guess different error types are fine then? have you spun it up locally and tried to trigger all of the errors (both the new Arithmetic error and the previous custom Error<T> errors)?

pallets/automation-time/src/tests_calculation.rs
justinzhou93: why specifically this one test - there's a bunch of overflow errors. are you just trying to test any one of them to make sure one succeeds? If so, to me, this one scenario seems pretty impossible to actually happen, since we define the MaxScheduleSeconds and we won't ever set it to MAX and the current_time_slot shouldn't really be incorrect, so this overflow should never realistically happen. is there a more realistic example that you could test? (i say realistic because it might be easier to replicate). So for example, https://github.com/OAK-Foundation/OAK-blockchain/pull/187/files#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R1154 or https://github.com/OAK-Foundation/OAK-blockchain/pull/187/files#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R1159 might be easier to mock without having to write a whole new runtime file

but if neither of those work and you can't find any other alternative than to remock the entire runtime file, is there any chance you might be able to write a runtime file without any of the XCM pieces? there's a lot of stuff in this mock runtime that may not necessarily need to be there i think? (or maybe it does, but see if you can rip stuff out since it shouldn't bother your test)

also, couldn't you test with validate_and_schedule_task function rather than the schedule notify extrinsic, since the validate_and_schedule_task function was the one changed?

irsal:
please pull latest master branch

pallets/automation-time/src/lib.rs
imstar15: The previous custom Error errors:

image

Arithmetic error:

image

pallets/automation-time/src/tests_calculation.rs
imstar15: **1. there's a bunch of overflow errors. are you just trying to test any one of them to make sure one succeeds?**

Yes, it's hard to construct the test conditions that trigger it since these cases rarely happen.
In fact this is the only test condition I have constructed.

**2. is there any chance you might be able to write a runtime file without any of the XCM pieces?** 

Currently it doesn't, Origin must implement cumulus_pallet_xcm::Origin.

```
impl pallet_automation_time::Config for Test {
...
	type Origin = Origin;
}
```

Complation error: 
```
error[E0277]: the trait bound `Result<cumulus_pallet_xcm::Origin, tests_calculation::Origin>: From<tests_calculation::Origin>` is not satisfied
   --> pallets/automation-time/src/tests_calculation.rs:134:16
    |
134 |     type Origin = Origin;
    |                   ^^^^^^ the trait `From<tests_calculation::Origin>` is not implemented for `Result<cumulus_pallet_xcm::Origin, tests_calculation::Origin>`
    ```




https://github.com/OAK-Foundation/OAK-blockchain/blob/9ec1400cb72650d8d86667f36cc3982eed2ef99d/pallets/automation-time/src/lib.rs#L228

imstar15:
> please pull latest master branch

Yes, Based on latest branch.

pallets/automation-time/src/tests_calculation.rs
imstar15: **3. couldn't you test with validate_and_schedule_task function rather than the schedule notify extrinsic?**
 
 Good advice! I re-implemented as you suggested.

justinzhou93:
hmmm wait, i wanted to test this locally to see if i could rip out some of the xcm configs in this test file to make it more manageable, but it doesn't seem to be passing for me. this is the error i get below both when i run cargo test and cargo build

error[E0432]: unresolved import `parachain_staking::AdditionalIssuance`
    --> /Users/justinzhou/Documents/OAK/ENG/OAK-blockchain/pallets/vesting/src/lib.rs:37:5
     |
  37 | use parachain_staking::AdditionalIssuance;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `AdditionalIssuance` in the root

imstar15:
> hmmm wait, i wanted to test this locally to see if i could rip out some of the xcm configs in this test file to make it more manageable, but it doesn't seem to be passing for me. this is the error i get below both when i run cargo test and cargo build
> 
> error[E0432]: unresolved import `parachain_staking::AdditionalIssuance` --> /Users/justinzhou/Documents/OAK/ENG/OAK-blockchain/pallets/vesting/src/lib.rs:37:5 | 37 | use parachain_staking::AdditionalIssuance; | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `AdditionalIssuance` in the root

Yes, it would be better if the association of xcm config could be removed.

justinzhou93:
wait, but i'm having trouble with compiling. is it compiling locally for you?

imstar15:
> wait, but i'm having trouble with compiling. is it compiling locally for you?

Yes, it compiles successfully.

justinzhou93 (approved):
compilation errors seem to be something wrong with my local. code looks good and good to merge for now, but i would like to see a refactor to remove some of the unnecessary runtime configs in tests_calculation.rs

imstar15:
> compilation errors seem to be something wrong with my local. code looks good and good to merge for now, but i would like to see a refactor to remove some of the unnecessary runtime configs in tests_calculation.rs

Yes, that's right!

PR #188, opened 2022-07-06
add fee rpc

justinzhou93:


runtime/neumann/src/lib.rs
```diff
@@ -1103,23 +1103,32 @@ impl_runtime_apis! {
 					let action_enum = pallet_automation_time::Action::Notify {
 						message: "get_fees".into()
 					};
-					AutomationTime::calculate_execution_fee(&action_enum, executions)
+					match AutomationTime::calculate_execution_fee(&action_enum, executions) {
```
justinzhou93, line 1106: charles changed calculate_execution_fee to return a [result wrapped balance](https://github.com/OAK-Foundation/OAK-blockchain/pull/188/commits/a3ff79880bdb50a4b0a655cc69331fa4232ecf55#diff-e6a13f7032a383e19b77d20043ca46041d94f5e49396812e2de1b2000c301e59R1141) because of checked_add vs the saturated_add request from the audit. so now i have to unwrap. i don't wanna use .unwrap() because if it errors, unwrap will panic and kill our RPC node, which we don't wanna happen. so i'm just returning 0 if somehow we end up with a number that overflows a u128. otherwise, it returns the correct balance. is this fine?

justinzhou93: also should never overflow, since execution_times is limited to u32. i took the largest possible value for u32 and input into execution_times and it seems to not overflow u128 on the result.

image

if we go higher than u32, it returns rpc error:
image

rhuttman (approved):

pallets/automation-time/rpc/src/lib.rs
```diff
@@ -111,23 +113,27 @@ where
...
 		let try_into_rpc_balance = |value: Balance| {
-			value.try_into().map_err(|_| RpcError {
-				code: ErrorCode::InvalidParams,
-				message: format!("RPC value doesn't fit in NumberOrHex representation"),
-				data: None,
+			value.try_into().map_err(|_| {
+				JsonRpseeError::Call(CallError::Custom(ErrorObject::owned(
+					Error::RuntimeError.into(),
+					"RPC value doesn't fit in NumberOrHex representation",
+					Some(format!("RPC value cannot be translated into NumberOrHex representation")),
```
justinzhou93, line 135: it doesn't like `None` because it can't serialize for some reason

runtime/neumann/src/lib.rs
```diff
@@ -1151,10 +1151,64 @@ impl_runtime_apis! {
...
+					let action_enum = pallet_automation_time::Action::XCMP {
+						para_id: cumulus_primitives_core::ParaId::from(2114),
+						call: vec![0, 1],
+						weight_at_most: 1_000_000_000,
+					};
+					match AutomationTime::calculate_execution_fee(&action_enum, executions) {
+						Ok(balance) => balance,
+						Err(_e) => 0,
+					}
+				},
```
atomaka, line 1206: Do we need to add a case for `Action::AutoCompoundDelegatedStake` now that https://github.com/OAK-Foundation/OAK-blockchain/pull/180 is merged?

pallets/automation-time/rpc/Cargo.toml
```diff
@@ -19,6 +19,7 @@ sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkad
 sp-api = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.23" }
 sp-blockchain = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.23" }
 sp-block-builder = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.23" }
+sp-rpc = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20" }
```
lreesby, line 22: should this be bumped to v0.9.23?

justinzhou93: oh lordie, thanks so much for the catch haha

node/Cargo.toml
```diff
@@ -46,6 +46,7 @@ jsonrpsee = { version = "0.13.0", features = ["server"] }
 neumann-runtime = { path = "../runtime/neumann", optional = true}
 turing-runtime = { path ='../runtime/turing', optional = true}
 pallet-automation-time-rpc ={ path = "../pallets/automation-time/rpc"}
+pallet-automation-time ={ path = "../pallets/automation-time"}
```
nvengal, line 49: Is this still required?

pallets/automation-time/rpc/src/lib.rs
```diff
@@ -112,6 +123,37 @@ where
...
+			api.get_time_automation_fees(&at, action, executions).map_err(|e| {
+				CallError::Custom(ErrorObject::owned(
+					Error::RuntimeError.into(),
+					"Unable to get time automation fees",
+					Some(e.to_string()),
+				))
+			})?;
+
+		let try_into_rpc_balance = |value: Balance| {
+			value.try_into().map_err(|_| {
```
nvengal, line 146: Was `NumberOrHex` a workaround for serialization with u128? What if you `try_into()` a `u64` instead? Since these are fees I expect they'll always be pretty small comparatively.

pallets/automation-time/rpc/src/lib.rs
nvengal: You can debug the error here if you want. Can be useful for printing stack trace (maybe more?)- `format!("{:?}", e)` where e is the arg to map_err

pallets/automation-time/rpc/src/lib.rs
nvengal: This is u64 max: `1,844,674,407 TUR` we're definitely safe.

justinzhou93: ahaha, kk sounds good

pallets/automation-time/rpc/src/lib.rs
justinzhou93: chatted offline, error was `<Balance as TryInto<u64>>::Error` doesn’t implement `std::fmt::Debug`. would have to reimplement TryInto<u64>::Error, so just leaving this be for now

pallets/automation-time/rpc/src/lib.rs
justinzhou93: image

does indeed look better this way!

pallets/automation-time/rpc/src/lib.rs
```diff
@@ -112,6 +122,37 @@ where
...
+				JsonRpseeError::Call(CallError::Custom(ErrorObject::owned(
+					Error::RuntimeError.into(),
+					"RPC value doesn't fit in u64 representation",
+					Some(format!("RPC value cannot be translated into NumberOrHex representation")),
```
nvengal, line 149: Left over NumberOrHex error message

pallets/automation-time/Cargo.toml
```diff
@@ -41,6 +41,9 @@ pallet-timestamp = { git = "https://github.com/paritytech/substrate", default-fe
 ## Moonbeam Dependencies
 pallet-parachain-staking = { git = "https://github.com/OAK-Foundation/moonbeam", default-features = false, branch = "oak-polkadot-v0.9.23" }
 
+## Local
+pallet-automation-time-rpc-runtime-api = { path = "./rpc/runtime-api", default-features = false }
```
nvengal, line 45: This is the only part of this pr that I'm not totally sure about. I think we're fine importing this here and the compiler should have yelled at us if we were doing something really bad, but std/wasm issues have surprised me in the past. I think we're fine here, but flagging in case anyone has a second opinion.

nvengal (approved):
💯

PR #192, opened 2022-07-07
automation benchmarks increase deposit

LauraReesby:


justinzhou93 (approved):

pallets/automation-time/src/benchmarking.rs
```diff
@@ -28,11 +28,20 @@ use crate::Pallet as AutomationTime;
 const SEED: u32 = 0;
 // existential deposit multiplier
 const ED_MULTIPLIER: u32 = 1_000;
+// ensure enough funds to execute tasks
+const DEPOSIT_MULTIPLIER: u32 = 100_000_000;
```
nvengal, line 32: Does this serve a different purpose than the `ED_MULTIPLIER`? It looks like in most places the DEPOSIT_MULTIPLIER is just applied along with the ED_MULTIPLIER.

lreesby: `ED_MULTIPLIER` is being used to create the transfer amount for testing auto transfer.  The `DEPOSIT_MULTIPLIER` should be more than the `ED_MULTIPLIER` to ensure there is enough funds for both transferring and paying fees.

PR #193, opened 2022-07-08
v0.9.23 Upgrade

rhuttman:


runtime/turing/src/xcm_config.rs
```diff
@@ -404,8 +404,8 @@ parameter_types! {
 }
 
 parameter_type_with_key! {
-	pub ParachainMinFee: |_location: MultiLocation| -> u128 {
-		u128::MAX
+	pub ParachainMinFee: |_location: MultiLocation| -> Option<u128> {
+		None
```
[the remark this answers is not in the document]
rhuttman: They changed how to distinguish if you don't want to support this use case: https://github.com/open-web3-stack/open-runtime-module-library/commit/b2461e85e0f2aa3b055ab1c72670220a8ca5a196#diff-5035482cf0976bcbc7ae6640411e64eb3867ee351dde4b87f2e9c57294763bbdL64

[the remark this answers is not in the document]
nvengal: Brilliant. Thanks for the clarification.

nvengal (approved):

node/src/command.rs
```diff
@@ -374,14 +384,13 @@ pub fn run() -> Result<()> {
 
 				let id = ParaId::from(para_id);
 				let parachain_account =
-					AccountIdConversion::<polkadot_primitives::v2::AccountId>::into_account(&id);
+					AccountIdConversion::<polkadot_primitives::v2::AccountId>::into_account_truncating(&id);
```
justinzhou93, line 387: did they stop using into_account for some reason? into_account_truncating can come up with duplicate accounts at times: https://paritytech.github.io/cumulus/cumulus_primitives_core/struct.ParaId.html#method.into_account_truncating

nvengal (approved):

justinzhou93 (approved):
1 question, otherwise lgtm!

node/src/command.rs
rhuttman: hmm let me look into it. I just copied the parity parachain template

rhuttman: From what I can tell this is safe because we are only use it for ParaIds and pallets. never for normal addresses

justinzhou93 (approved):

PR #195, opened 2022-07-12
Return optimal autostaking period via an RPC

atomaka:
RPC to return the optimal period to auto-stake rewards.  Based on formula outlined in https://docs.google.com/document/d/1SNgeEXT7dcnQWucT43CKfaHrGHZQtfpER7YaNiyTTiM/edit# and documented at https://github.com/OAK-Foundation/OAK-blockchain/wiki/Auto-Compound-Staking-Rewards-API#querying-for-an-optimal-auto-compounding-time

Example call:

```
curl --location \
  --request POST \
  --data-raw '{"id":1, "jsonrpc":"2.0", "method": "automationTime_calculateOptimalAutostaking", "params": [50000000000000, "69NiyRvjQtQEGcQesiZgTyzHndUeGXZAaHYNgBGQtzhUg6R7"]}' \
 'http://localhost:8855' 
```

results in

```
{"jsonrpc":"2.0","result":{"period":46,"apy":0.14510475124271333},"id":1} 
```

Some additional testing for bad data:

```
Bad account prefix
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid SS58 prefix byte."},"id":1}
Bad account with good prefix
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid checksum"},"id":1}
Non-collating wallet
{"jsonrpc":"2.0","error":{"code":1,"message":"Unable to calculate optimal autostaking","data":"collator does not exist"},"id":1}
```


pallets/automation-time/src/autocompounding.rs
```diff
@@ -0,0 +1,221 @@
...
+use sp_std::{vec, vec::Vec};
+
+#[derive(Clone)]
+pub struct IntervalRow {
+	interval_earnings_after_fee: i128,
+	stake: i128,
+	ownership: f64,
+}
+
+pub fn do_calculate_optimal_autostaking(
```
atomaka, line 26: Type mess follows.

runtime/neumann/src/lib.rs
```diff
@@ -1155,6 +1156,30 @@ impl_runtime_apis! {
 		fn generate_task_id(account_id: AccountId, provided_id: Vec<u8>) -> Hash {
 			AutomationTime::generate_task_id(account_id, provided_id)
 		}
+
+		fn calculate_optimal_autostaking(principal: i128, collator: AccountId) -> AutostakingResult {
+			let candidate_info = ParachainStaking::candidate_info(collator);
+			let money_supply = Balances::total_issuance() + Vesting::total_unvested_allocation();
+
+			let collator_stake = candidate_info.unwrap().total_counted as i128;
```
atomaka, line 1164: This might not be a collator. Currently we return some dirty WASM error.  Should this method return a result to provider a cleaner error ("collator does not exist")?

runtime/neumann/src/lib.rs
```diff
@@ -1155,6 +1156,30 @@ impl_runtime_apis! {
...
+			let fee = (1 * DOLLAR) as i128;
```
atomaka, line 1165: Pending @nvengal 's PR

pallets/automation-time/src/autocompounding.rs
```diff
@@ -0,0 +1,221 @@
...
+	use super::*;
+
+	const DOLLAR: i128 = 10_000_000_000;
+
+	const AVERAGE_STAKING_DURATION: i32 = 90;
+	const ANNUAL_REWARDS_PER_COLLATOR: i128 = 1_041_667 * DOLLAR;
+	const DAILY_COLLATOR_AWARDS: i128 = ANNUAL_REWARDS_PER_COLLATOR / 365;
+
+	#[test]
+	fn test_1() {
```
justinzhou93, line 86: nit, any chance we might be able to put comment or more descriptive test name? sometimes it's hard to understand what each of the tests are testing for, since the difference seems to be just the amounts

runtime/neumann/src/lib.rs
```diff
@@ -1155,6 +1156,30 @@ impl_runtime_apis! {
...
+			let duration = 90;
```
justinzhou93, line 1166: 90 seconds or blocks or rounds or something else?

runtime/neumann/src/lib.rs
```diff
@@ -1155,6 +1156,30 @@ impl_runtime_apis! {
...
+			// 24 = number of collators
+			let daily_collator_rewards = (money_supply as f64 * 0.025) as i128 / 24 / 365;
```
justinzhou93, line 1168: should it be i128 or u128?

pallets/automation-time/src/autocompounding.rs
justinzhou93, line 26: if i'm reading this function correctly, i _think_ what's happening is:
1. the inner loop calculates the APY you earn with the various frequencies at which you auto compound, given a set duration you want to hold for. you store these APY values into a table.
2. the outer loop tries different durations and feeds those durations into the inner loop to calculate a "best" autocompounding frequency
3. you save the best of the best to get the best duration and frequency

is this the right interpretation?

justinzhou93 (approved):
nits, but otherwise lgtm!

runtime/neumann/src/lib.rs
```diff
@@ -1155,6 +1156,30 @@ impl_runtime_apis! {
 		fn generate_task_id(account_id: AccountId, provided_id: Vec<u8>) -> Hash {
 			AutomationTime::generate_task_id(account_id, provided_id)
 		}
+
+		fn calculate_optimal_autostaking(principal: i128, collator: AccountId) -> AutostakingResult {
+			let candidate_info = ParachainStaking::candidate_info(collator);
+			let money_supply = Balances::total_issuance() + Vesting::total_unvested_allocation();
```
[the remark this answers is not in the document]
nvengal: yeah, for your purpose. Money supply is calculated the same way.

runtime/neumann/src/lib.rs
nvengal: Is this an inadvertent truncation or is it on purpose?

runtime/neumann/src/lib.rs
```diff
@@ -1156,6 +1157,34 @@ impl_runtime_apis! {
...
+
+			let collator_stake = match candidate_info {
+				Some(x) => Ok(x.total_counted as i128),
+				None => Err("collator does not exist"),
+			}?;
+
+			let fee = (1 * DOLLAR) as i128;
+			let duration = 90;
+			// 24 = number of collators
```
nvengal, line 1175: This is technically a dynamic value: `ParachainStaking::total_selected()`

nvengal (commented):
Nothing blocking but some questions/suggestions. Turns out, I don't know how to do math properly either so sadly lacking any feedback on the types there.

runtime/neumann/src/lib.rs
atomaka: Discussed in Slack. Dynamic as in we configure it to a static number and not based on round. We definitely want to use this.

pallets/automation-time/src/autocompounding.rs
atomaka: Discussed in slack. These are arbitrary tests generated to make sure changing different inputs results in the expected output and have no meaningful purpose. I added a comment to note this.

runtime/neumann/src/lib.rs
atomaka: It is intentional. The formula was originally calculated with static integer values so I'm happy with this truncation.  There are many other more offensive truncations in the called method though xD

runtime/neumann/src/lib.rs
atomaka: 90 days.

nvengal (approved):

PR #196, opened 2022-07-14
Add multisig pallet

imstar15:
### Create multisig wallet

Click the multisig button in the account UI

image

Please select the participants(Order insensitive) of the multisig wallet and set the threshold.

image

You get a multisig wallet.

image

### Authorize transaction

When you send extrinsic using a multisig wallet, you need to authorize it.

**Note: Be sure to save the multisig call data.**

image

Another participant will receive multisig approval on the UI. He can approve this approval.

image

Paste multisig call data on this UI.

image

image

When the number of Approves reaches a threshold, this extrinsic is executed on-chain.

### DIFF
--- a/runtime/turing/src/lib.rs
+++ b/runtime/turing/src/lib.rs
@@ -44,8 +44,8 @@ use sp_version::RuntimeVersion;
 use frame_support::{
 	construct_runtime, parameter_types,
 	traits::{
-		ConstU128, ConstU32, Contains, EnsureOneOf, Imbalance, InstanceFilter, OnUnbalanced,
-		PrivilegeCmp,
+		ConstU128, ConstU16, ConstU32, Contains, EnsureOneOf, Imbalance, InstanceFilter,
+		OnUnbalanced, PrivilegeCmp,
 	},
 	weights::{
 		constants::{BlockExecutionWeight, ExtrinsicBaseWeight, RocksDbWeight, WEIGHT_PER_SECOND},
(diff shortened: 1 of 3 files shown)


irsal:
pull latest please

imstar15:
> pull latest please

Yes, rebased on latest.

runtime/turing/src/lib.rs
```diff
@@ -288,6 +288,23 @@ impl frame_system::Config for Runtime {
 	type MaxConsumers = ConstU32<16>;
 }
 
+parameter_types! {
+	// One storage item; key size is 32; value is size 4+4+16+32 bytes = 56 bytes.
```
nvengal, line 292: Trying to understand where these numbers are from. My reading of the docs come out differently.
```
/// This is held for an additional storage item whose value size is
/// `4 + sizeof((BlockNumber, Balance, AccountId))` bytes and whose key size is
/// `32 + sizeof(AccountId)` bytes.
```
https://github.com/paritytech/substrate/blob/4a19d408095ebcb47840ad1f2f086f53a0020cb7/frame/multisig/src/lib.rs#L134-L136

4 + sizeof((BlockNumber, Balance, AccountId)) = 4 + sizeof((u32, u128, u32(?))) = 4 + 4 + 16 + 4 = 28 bytes
32 + sizeof(AccountId) = 32 + 4 = 36 bytes
total = 28 + 36 = 64 bytes

imstar15: OK, I'll go look up some sources and understand it.
If I figure it out, I'll reply.

atomaka (approved):
👍

PR #200, opened 2022-07-19
Add rpc endpoint to retrieve auto compound task ids

nvengal:


runtime/neumann/src/lib.rs
```diff
@@ -1191,6 +1191,21 @@ impl_runtime_apis! {
...
+				.map_or(vec![], |s| s.delegations.0).into_iter()
+				.map(|d| d.owner)
+				.map(|collator_id| {
+					AutomationTime::generate_auto_compound_delegated_stake_provided_id(&account_id, &collator_id)
+				})
+				.map(|provided_id| {
+					AutomationTime::generate_task_id(account_id.clone(), provided_id)
+				})
+				.filter(|task_id| {
+					AutomationTime::get_task(task_id).is_some()
```
nvengal, line 1206: Originally this was going to return a `task` struct, but I need to spend some more time understanding serializing before I can do that. For now, this should enable dApp developers to get the tasks directly from storage. Eventually we can serialize the task rather than just filtering here.

PR #201, opened 2022-07-19
Prevent frequency of 0 for reoccurring task

nvengal:


atomaka (approved):

justinzhou93 (approved):
whoops...sorry i missed this first time around!

pallets/automation-time/src/lib.rs
```diff
@@ -606,12 +606,14 @@ pub mod pallet {
...
 				Self::generate_auto_compound_delegated_stake_provided_id(&who, &collator_id);
-			if frequency % 3600 != 0 {
-				Err(Error::<T>::InvalidTime)?
-			}
-			if frequency > T::MaxScheduleSeconds::get() {
-				Err(Error::<T>::TimeTooFarOut)?
+
+			let next_execution =
+				execution_time.checked_add(frequency).ok_or(Error::<T>::TimeTooFarOut)?;
+			Self::is_valid_time(next_execution)?;
```
lreesby, line 612: If I'm reading correctly you did it this way to both validate the `execution_time` and `frequency` at the same time?  You are ensuring `frequency` is not `0` and also that the `execution_time` is valid (`% 3600`).  It might be helpful to add a comment what you are trying to accomplish.  Might not be clear to someone in the future.

pallets/automation-time/src/lib.rs
```diff
@@ -606,12 +606,14 @@ pub mod pallet {
...
+			if next_execution == execution_time {
```
lreesby, line 613: could `frequency` ever be negative?  any need to check `next_execution <= execution_time`?

nvengal: frequency is a `u64` so we're protected there

pallets/automation-time/src/lib.rs
nvengal: Execution time is validated later on. This is to validate frequency without having to rewrite the valid_time code. Added a comment to that effect.

lreesby (approved):
lgtm

PR #204, opened 2022-07-19
Upgrade and lock in srtool version

rhuttman:


atomaka (approved):
1.62.0 is the current tag so locking it doesn't change anything in the test.  We're just changing srtool version here. Seems fine to test this way.

PR #206, opened 2022-07-20
Using const more consistently in runtime

rhuttman:


justinzhou93 (approved):

runtime/neumann/src/lib.rs
```diff
@@ -595,8 +589,6 @@ impl cumulus_pallet_aura_ext::Config for Runtime {}
 
 parameter_types! {
 	pub const Period: u32 = 6 * HOURS;
-	pub const Offset: u32 = 0;
```
nvengal, line 598: Looks like `Offset` was just unused? Surprised that we didn't see a compiler warning.

nvengal (approved):

runtime/neumann/src/lib.rs
rhuttman: ya it made me wonder what else is just unused that the compiler missed

PR #208, opened 2022-07-20
Use saturating math for fees

nvengal:
Rather than throwing an error while calculating execution fees we should saturate the value at the weight and balance boundaries. Not only is this exceedingly unlikely to happen in real world scenarios, the saturated value provides a reasonable unreasonable fee in the extraordinary case.


pallets/automation-time/src/tests_calculation.rs
```diff
@@ -1,213 +0,0 @@
...
-fn new_test_ext(state_block_time: u64) -> sp_io::TestExternalities {
-	let t = system::GenesisConfig::default().build_storage::<Test>().unwrap();
-	let mut ext = sp_io::TestExternalities::new(t);
-	ext.execute_with(|| System::set_block_number(1));
-	ext.execute_with(|| Timestamp::set_timestamp(state_block_time));
-	ext
-}
-
-#[test]
-fn validate_and_schedule_task_arithmetic_overflow() {
```
nvengal, line 200: This test is not providing much value while requiring a decent amount of upkeep when the pallet changes (due to the amount of repeated code in this file). I've opted to just remove it. Thoughts?

lreesby (approved):
lgtm

justinzhou93:
i don't mind reverting these changes, but i believe that charles made them in accordance to the audit report. I do think that it's pretty fair that the likelihood of the math going wrong is very very low, but how much do we care about not fixing something that the audit report called out? is that bad in the eyes of the community or something like that?

atomaka:
> i don't mind reverting these changes, but i believe that charles made them in accordance to the audit report. I do think that it's pretty fair that the likelihood of the math going wrong is very very low, but how much do we care about not fixing something that the audit report called out? is that bad in the eyes of the community or something like that?

I don't think we're reverting anything here. One of our adds is being swapped from a `checked_add` to a `saturating_add`.  Both operations should be equally secure but have a slightly different overflow behavior. `checked_add` will panic and `saturating_add` will return the max value for the type. So we've still prevented overflow, just in a different way.

pallets/automation-time/src/lib.rs
```diff
@@ -1431,31 +1424,27 @@ pub mod pallet {
...
 
-		pub fn calculate_execution_fee(
-			action: &Action<T>,
-			executions: u32,
-		) -> Result<BalanceOf<T>, DispatchError> {
+		/// Calculates the execution fee for a given action based on weight and num of executions
+		///
+		/// Fee saturates at Weight/BalanceOf when there are an unreasonable num of executions
+		/// In practice, executions is bounded by T::MaxExecutionTimes and unlikely to saturate
+		pub fn calculate_execution_fee(action: &Action<T>, executions: u32) -> BalanceOf<T> {
```
atomaka, line 1431: 👍

runtime/turing/src/lib.rs
```diff
@@ -1209,14 +1206,7 @@ impl_runtime_apis! {
...
 				candidate_info.ok_or("collator does not exist")?.total_counted as i128;
-			let fake_action = pallet_automation_time::Action::AutoCompoundDelegatedStake {
-				delegator: sp_runtime::AccountId32::new([1u8; 32]),
-				collator: sp_runtime::AccountId32::new([2u8; 32]),
-				account_minimum: 1_000_000_000,
-				frequency: 60 * 60 * 24 * 90,
-			};
-			let fee = AutomationTime::calculate_execution_fee(&fake_action, 1)
-				.map_err(|_| "could not calculate fee")? as i128;
+			let fee = AutomationTime::calculate_execution_fee(&(AutomationAction::AutoCompoundDelegatedStake.into()), 1) as i128;
```
atomaka, line 1209: Neat.

atomaka (approved):

justinzhou93:
> > i don't mind reverting these changes, but i believe that charles made them in accordance to the audit report. I do think that it's pretty fair that the likelihood of the math going wrong is very very low, but how much do we care about not fixing something that the audit report called out? is that bad in the eyes of the community or something like that?
> 
> I don't think we're reverting anything here. One of our adds is being swapped from a `checked_add` to a `saturating_add`. Both operations should be equally secure but have a slightly different overflow behavior. `checked_add` will panic and `saturating_add` will return the max value for the type. So we've still prevented overflow, just in a different way.

ahh i think we used to have saturating_add here, which is what i meant by saying "reverting". This was changed to checked_add for the purposes of fixing a "vulnerability" found by slow mist (see picture below).
image

I do tend to agree that this is a bit of a stretch to say that it ever does overflow, but slow mist's concern was that if it should overflow, they would prefer an explicit error rather than the maxed value, which would be an incorrect value. i.e. they would rather it be clear to the user that an arithmetic error has occurred rather than giving them a value that is incorrect. so i think my question is more, whether we are okay in disregarding this advice from Slow Mist. It is just of threat level "Suggestion", which is the lowest level, so i think it's fine? but 🤷 , figured i'd bring it up just in case

justinzhou93 (approved):

justinzhou93:
chatted with andrew on slack about this, but mainly the concern here is a public perception issue that we didn't follow through with suggestion from audit. from a technical perspective, no issues at all, since overflowing a u128 for fees is pretty much impossible. 

but after conversation, it's prob fair to look at audit as more of a thought process rather than list of rules to follow. so, maybe just a conversation with growth to ensure they're ready for the question should that question ever arise. otherwise, lgtm, so no blocker here!

atomaka:
> ahh i think we used to have saturating_add here, which is what i meant by saying "reverting". This was changed to checked_add for the purposes of fixing a "vulnerability" found by slow mist (see picture below). 

Ignore me. Totally a revert xD.

nvengal:
Added this comment to function as some level of documentation: https://github.com/OAK-Foundation/OAK-blockchain/blob/1c004dde77341e7250d59554f9fe4146e251d054/pallets/automation-time/src/lib.rs#L1427-L1430
Will check if I should be documenting anywhere else wrt the audit.

PR #211, opened 2022-07-22
Remove tight coupling of valve + automation-time config

nvengal:
Swaps the explicit coupling of valve and automation-time for an implicit coupling.
Cons:
- Still implicitly coupled due to events and extrinsic expecting automation-time

Pros:
- Delete ~300 lines of test boilerplate

Local testing:
Task was added directly onto queue and does not run until valve is reopened
image


pallets/valve/src/lib.rs
```diff
@@ -46,18 +49,20 @@ pub mod pallet {
...
 		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
 		/// Weight information for the extrinsics in this module.
 		type WeightInfo: WeightInfo;
 
 		/// The pallets that we want to close on demand.
 		type ClosedCallFilter: Contains<Self::Call>;
+
+		/// The AutomationTime pallet.
+		type AutomationTime: Shutdown;
```
nvengal, line 65: The valve pallet makes many assumptions about the automation time pallet. I've only unraveled some of those assumptions here (events and extrinsics are still about tasks start/stopping, etc) so for now I've tried to make it clear that this type is expecting the automation time pallet.

It's not great but I think it's still a step in the right direction of decoupling these pallets completely.

pallets/valve/src/mock.rs
```diff
@@ -267,38 +87,35 @@ impl Contains<Call> for ClosedCallFilter {
...
-	fn close_pallet_gate_existing() -> Weight {
-		0
-	}
-	fn open_pallet_gate() -> Weight {
-		0
-	}
-	fn open_pallet_gates() -> Weight {
-		0
+thread_local! {
+	pub static MOCK_AUTOMATION_TIME_STORE: RefCell<bool> = RefCell::new(false);
```
nvengal, line 91: I picked up this pattern for mocking in tests from the Session pallet. Big improvement over having to stub the entire pallet.

pallets/automation-time/src/lib.rs
```diff
@@ -1447,4 +1447,16 @@ pub mod pallet {
 			T::ExecutionWeightFee::get().saturating_mul(weight_as_balance)
 		}
 	}
+
+	impl<T: Config> pallet_valve::Shutdown for Pallet<T> {
+		fn is_shutdown() -> bool {
+			Pallet::<T>::is_shutdown()
```
justinzhou93, line 1453: is it Pallet::<T>::is_shutdown() or is it Self::is_shutdown? most of our other functions to get DB storage is querying the function off of `Self::` rather than `Pallet::<T>::`. although could just be the same syntax, prob good to keep consistency

justinzhou93 (approved):
1 nitpick about consistency on calling DB GET functions, but otherwise, lgtm!

justinzhou93 (approved):

rhuttman (approved):

PR #216, opened 2022-07-28
Add RPC to return Turing cross chain account

atomaka:
Test RPC with

```
echo "Attempting to get cross chain account for Alice"
curl --location \
  --request POST \
  --data-raw '{"id":1, "jsonrpc":"2.0", "method": "xcmpHandler_crossChainAccount", "params": ["5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"]}' \
 'http://localhost:8855'
echo
echo "SUCCESS" 
```


pallets/xcmp-handler/rpc/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+#[async_trait]
+impl<C, Block> XcmpHandlerApiServer<<Block as BlockT>::Hash> for XcmpHandler<C, Block>
+where
+	Block: BlockT,
+	C: Send + Sync + 'static + ProvideRuntimeApi<Block> + HeaderBackend<Block>,
+{
+	fn cross_chain_account(&self, account_id: AccountId32) -> RpcResult<AccountId32> {
+		let multiloc = MultiLocation::new(
+			1,
+			X2(Parachain(2114), Junction::AccountId32 { network: Any, id: account_id.into() }),
```
atomaka, line 82: Is there anyway to get our parachain id outside of the runtime?

rhuttman: You could point to the chain specs, but then they need to tell you what chain they are calling for. You can call the runtime from here to get the correct Id though

atomaka:
Closed in favor of #222

PR #220, opened 2022-08-02
Fix XCM fee calculation

rhuttman:


pallets/xcmp-handler/src/tests.rs
```diff
@@ -157,14 +158,14 @@ fn returns_fee() {
 }
 
 #[test]
-fn errors_if_not_overflow() {
```
justinzhou93, line 160: i'm fine with us changing the test, but just wanna make sure it's not unintentional. do we still need a case where xcm fee overflows somehow?

justinzhou93 (approved):

pallets/xcmp-handler/src/tests.rs
rhuttman: that still causes an overflow

justinzhou93: ahh sorry sorry, i misread test name. ignore me

PR #222, opened 2022-08-02
Provide RPCs to support XCMP

atomaka:
Account RPC:

Parameter is account id. (included is Alice's account)

```
curl --location \
  --request POST \
  --data-raw '{"id":1, "jsonrpc":"2.0", "method": "xcmpHandler_crossChainAccount", "params": ["5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"]}' \
 'http://localhost:8855' 
```

Fee RPC:

Parameter is a signed extrinsic.  (included is signed scheduled remarkWithEvent from Alice)

```
curl --location \
  --request POST \
  --data-raw '{"id":1, "jsonrpc":"2.0", "method": "xcmpHandler_fees", "params":
  ["0x41028400d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d0166059db25528559f4bc8ac315de7f321fb9c492a5f0e6ebe13fdbea410695972a0168053cce11f61a7992bbd5394f4d45ece007f6190e43302447293c9129b85940004003c02041004000000000000000042080000003800082c48656c6c6f20776f726c64808d5b0000000000"]}' \
 'http://localhost:8855' 
```


pallets/xcmp-handler/rpc/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+xcm = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.23" }
+xcm-executor = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.23" }
+xcm-builder = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "release-v0.9.23" }
+
+pallet-xcmp-handler-rpc-runtime-api = { path = "./runtime-api" }
+
+[features]
+default = ["std"]
+std = [
```
rhuttman, line 33: why don't the other deps need to be there?

runtime/turing/src/lib.rs
```diff
@@ -1180,6 +1197,39 @@ impl_runtime_apis! {
...
+			if let Call::AutomationTime(pallet_automation_time::Call::schedule_xcmp_task{
+				provided_id, execution_times, para_id, currency_id, encoded_call, encoded_call_weight
+			}) = extrinsic.clone().function {
+				let len = encoded_xt.len() as u32;
+
+				let xcm_fee = XcmpHandler::calculate_xcm_fee_and_weight(
+					u32::from(para_id),
+					CurrencyId::from(currency_id),
+					encoded_call_weight,
+				).unwrap().0;
```
rhuttman, line 1216: is this the right way to handle potential errors for RPCs?

atomaka: @rhuttman Definitely not and I was hoping you had a better solution.  My last RPC implemented resolved this in a more user friendly way (https://github.com/OAK-Foundation/OAK-blockchain/blob/master/runtime/turing/src/lib.rs#L1206). It didn't feel like a very good approach either.

pallets/xcmp-handler/rpc/Cargo.toml
atomaka: The better question is why does this dependency need to be here.  Since this code is only executed on node, we don't need include the `std` feature (note automation time's RPC does not include the feature).  So why did I add it? `xcm-executor` is expecting the `sp_std::vec` macro to be present.

```
error: cannot find macro `vec` in this scope
   --> /Users/atomaka/.cargo/git/checkouts/polkadot-4038f27d5e4ea2e8/a7e188c/xcm/xcm-executor/src/lib.rs:177:23
    |
177 |             error_handler: Xcm(vec![]),
    |                                ^^^
    |
    = note: consider importing one of these items:
            parity_scale_codec::alloc::vec
            sp_std::vec
```

runtime/turing/src/lib.rs
rhuttman: I would like a similar thing here is correct. If we do not support that chain/currency combo then we should ultimately return an error. I would think this fcn returns a result and the RPC fcn that calls this wraps that into a runtime error w/ a special message

pallets/xcmp-handler/rpc/Cargo.toml
rhuttman: that might be an argument to move that logic into the runtime, but I am fine either way

pallets/xcmp-handler/rpc/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+					)))
+				})
+			})
+			.map_err(|_| {
+				JsonRpseeError::Call(CallError::Custom(ErrorObject::owned(
+					Error::RuntimeError.into(),
+					"RPC value doesn't fit in u64 representation",
+					Some(format!("RPC value cannot be translated into u64 representation")),
+				)))
+			})?
```
atomaka, line 118: This is bad

rhuttman (approved):
good enough to unblock

pallets/xcmp-handler/rpc/src/lib.rs
rhuttman: agreed but we can circle back tomorrow. Just want to get mangata unblocked

PR #225, opened 2022-08-08
add price triggers demo

justinzhou93:


pallets/automation-price/src/fees.rs
```diff
@@ -0,0 +1,83 @@
+// This file is part of OAK Blockchain.
```
justinzhou93, line 1: copy pasted from automation time

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,700 @@
...
+		#[transactional]
+		pub fn schedule_swap_task(
+			origin: OriginFor<T>,
+			provided_id: Vec<u8>,
+			asset: Vec<u8>,
+			direction: u8,
+			trigger_percentage: u128,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			let fee = <BalanceOf<T>>::saturated_from(1_000_000_000_000u64);
```
justinzhou93, line 301: i figured i'd make it expensive, lemme know if it's too expensive (it's like 1-1.5 TUR)

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,700 @@
...
+			value: u128,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			if let Some(asset_metadatum) = Self::get_asset_metadata(asset.clone()) {
+				let asset_sudo: AccountOf<T> = asset_metadatum.asset_sudo;
+				if asset_sudo != who {
+					Err(Error::<T>::InvalidAssetSudo)?
+				}
+			}
+			let fee = <BalanceOf<T>>::saturated_from(1_000_000_000_000u64);
```
justinzhou93, line 377: same here, very expensive

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,700 @@
...
+	impl<T: Config> TaskHashInput<T> {
+		pub fn create_hash_input(owner_id: AccountOf<T>, provided_id: Vec<u8>) -> TaskHashInput<T> {
+			TaskHashInput::<T> { owner_id, provided_id }
+		}
+	}
+
+	#[derive(Debug, Encode, Decode, TypeInfo)]
+	#[scale_info(skip_type_params(T))]
+	pub struct AssetMetadatum<T: Config> {
+		upper_bound: u16,
```
justinzhou93, line 121: upper_bound and lower_bound aren't currently used. but the idea is that we have limited slots (currently 20). so you can define what percentiles increase/decrease those slots would be. currently it's 1% per slot, but upper bound of like 100 would mean 5% increase per slot in the upward direction. decreases will be governed by lower bound

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,700 @@
...
+		/// * `directions`: number of directions of data input. (up, down, ?)
+		///
+		/// # Errors
+		#[pallet::weight(<T as Config>::WeightInfo::delete_asset_extrinsic())]
+		#[transactional]
+		pub fn delete_asset(
+			origin: OriginFor<T>,
+			asset: Vec<u8>,
+		) -> DispatchResult {
+			// TODO: needs fees if opened up to non-sudo
```
justinzhou93, line 416: currently doesn't have fees

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,700 @@
...
+				// delete assets' tasks
+				let asset_reset_weight = <T as Config>::WeightInfo::reset_asset(scheduled_deletion_assets.len().try_into().unwrap());
+				if weight_left < asset_reset_weight {
+					return weight_left
+				}
+				// TODO: this assumes that all assets that need to be reset in a period can all be done successfully in a block.
+				// 			 in the future, we need to make sure to be able to break out of for loop if out of weight and continue 
+				//       in the next block. Right now, we will not run out of weight - we will simply not execute anything if 
+				//       not all of the asset resets can be run at once. this may cause the asset reset triggers to not go off,
+				//       but at least it should not brick the chain.
```
justinzhou93, line 457: pls read and lemme know if this is ok!!!

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,700 @@
...
+			};
+		}
+
+		pub fn get_current_time_slot() -> Result<UnixTime, Error<T>> {
+			let now = <timestamp::Pallet<T>>::get().saturated_into::<UnixTime>();
+			if now == 0 {
+				Err(Error::<T>::BlockTimeNotSet)?
+			}
+			let now = now.saturating_div(1000);
+			let diff_to_min = now % 60;
```
justinzhou93, line 504: check once a minute? or too frequent?

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,700 @@
...
+			Ok(now.saturating_sub(diff_to_min))
+		}
+
+		pub fn delete_asset_tasks(asset: Vec<u8>) -> Weight {
+			// delete scheduled tasks
+			ScheduledTasks::<T>::remove_prefix((asset.clone(), ), None);
+			// delete tasks from tasks table
+			Tasks::<T>::remove_prefix((asset.clone(), ), None);
```
justinzhou93, line 512: this _should_ be 1 write, as per moonbeam. e.g. https://github.com/PureStake/moonbeam/commit/2a5e7e82e01d187ea547c91e51df653f2c7047cb#diff-4568d58ba37a8c54256c2853ef41bdfd28bbd61eb4c5c57b75a8787dd1164d2cR177 (the read is from the if statement above this line)

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,699 @@
...
+			}
+
+			// remove assets as necessary
+			let current_time_slot = match Self::get_current_time_slot() {
+				Ok(time_slot) => time_slot,
+				Err(_) => return weight_left,
+			};
+			if let Some(scheduled_deletion_assets) = Self::get_scheduled_asset_period_reset(current_time_slot) {
+				// delete assets' tasks
+				let asset_reset_weight = <T as Config>::WeightInfo::reset_asset(scheduled_deletion_assets.len().try_into().unwrap());
```
justinzhou93, line 449: no weight calc for delete_asset_tasks or update_asset_reset because it's wrapped up in here

pallets/automation-price/src/weights.rs
```diff
@@ -0,0 +1,93 @@
...
+	fn delete_asset_extrinsic() -> Weight{
+		(220_000_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(2 as Weight))
+			.saturating_add(T::DbWeight::get().writes(6 as Weight))
+	}
+	fn asset_update_extrinsic() -> Weight{
+		(220_000_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(3 as Weight))
+			.saturating_add(T::DbWeight::get().writes(1 as Weight))
+			.saturating_add(T::DbWeight::get().reads(21 as Weight))
```
justinzhou93, line 80: this is a ridiculous number because of https://github.com/OAK-Foundation/OAK-blockchain/pull/225/files#diff-1040044567b08cfbb340dcaa8af68f80a19a1ab98db9c94831417c4d982d6eccR658. since the amount that the price increases dictates how much work the chain will do, we have to do the maximum amount. that's why i'm also hesitant to give more than 20 buckets here, so we're currently capped at 20% increase/decreases in asset value

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,699 @@
...
+		) -> DispatchResult {
+			let mut existing_task_queue: Vec<(Vec<u8>, T::Hash)> = Self::get_task_queue();
+			// TODO: fix adjusted_higher to not peg to 20. Should move with the removal of 100 % increase.
+			let adjusted_higher = match higher > 20 {
+				true => 20,
+				false => higher,
+			};
+			for percentage in lower..adjusted_higher {
+				// TODO: pull all and cycle through in memory
+				if let Some(asset_tasks) = Self::get_scheduled_tasks((asset.clone(), direction.clone(), percentage.clone())) {
```
justinzhou93, line 664: this is a lot of reads. ideally i figure out how to pull into memory and cycle through and reinsert large parts of the storage map. but idk how to do that right now

pallets/automation-price/src/lib.rs
justinzhou93: also, is this a safe unwrap? i think so? scheduled_deletion_assets is usize where as reset_asset is expecting a u64? so it should be fine to unwrap, right? i doubt there will be more than 2^64 assets in our DB 😅

justinzhou93 (commented):
first cut, did QA, but not quite stress test. prob good for first cut of review. emphasis on making sure things don't brick the chain. if curious, also feel free to spin up and play around to make sure i didn't miss too many edge cases. mainly concerned with accidentally bricking chain through some sort of edge case.

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+			mut task_ids: Vec<(Vec<u8>, T::Hash)>,
+			mut weight_left: Weight,
+		) -> (Vec<(Vec<u8>, T::Hash)>, Weight) {
+			let mut consumed_task_index: usize = 0;
+			for task_id in task_ids.iter() {
+				consumed_task_index.saturating_inc();
+				let action_weight = match Self::get_task(task_id) {
+					None => {
+						Self::deposit_event(Event::TaskNotFound { task_id: task_id.1.clone() });
+						<T as Config>::WeightInfo::emit_event()
```
rhuttman, line 608: is this also a read?

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+			}
+
+			// run as many scheduled tasks as we can
+			let task_queue = Self::get_task_queue();
+			weight_left = weight_left.saturating_sub(T::DbWeight::get().reads(1 as Weight));
+			if task_queue.len() > 0 {
+				let (tasks_left, new_weight_left) = Self::run_tasks(task_queue, weight_left);
+				TaskQueue::<T>::put(tasks_left);
+				weight_left =
+					new_weight_left.saturating_sub(T::DbWeight::get().writes(1 as Weight));
```
rhuttman, line 512: needs to be factored into the `run_tasks` call

pallets/automation-price/src/lib.rs
```diff
@@ -0,0 +1,793 @@
...
+			let current_time_slot = match Self::get_current_time_slot() {
+				Ok(time_slot) => time_slot,
+				Err(_) => return weight_left,
+			};
+			if let Some(scheduled_deletion_assets) =
+				Self::get_scheduled_asset_period_reset(current_time_slot)
+			{
+				// delete assets' tasks
+				let asset_reset_weight = <T as Config>::WeightInfo::reset_asset(
+					scheduled_deletion_assets.len().try_into().unwrap(),
```
rhuttman, line 483: can saturate into

runtime/neumann/src/lib.rs
```diff
@@ -1049,6 +1063,7 @@ construct_runtime!(
 		AutomationTime: pallet_automation_time::{Pallet, Call, Storage, Event<T>} = 60,
 		Vesting: pallet_vesting::{Pallet, Storage, Config<T>, Event<T>} = 61,
 		XcmpHandler: pallet_xcmp_handler::{Pallet, Call, Storage, Event<T>} = 62,
+		AutomationPrice: pallet_automation_price::{Pallet, Call, Storage, Event<T>} = 63,
```
rhuttman, line 1066: make some absurd number

runtime/turing/src/lib.rs
```diff
@@ -1109,6 +1123,7 @@ construct_runtime!(
 		AutomationTime: pallet_automation_time::{Pallet, Call, Storage, Event<T>} = 60,
 		Vesting: pallet_vesting::{Pallet, Storage, Config<T>, Event<T>} = 61,
 		XcmpHandler: pallet_xcmp_handler::{Pallet, Call, Storage, Event<T>} = 62,
+		AutomationPrice: pallet_automation_price::{Pallet, Call, Storage, Event<T>} = 63,
```
rhuttman, line 1126: make absurd number

rhuttman (changes requested):

justinzhou93:
note to self: 
- make asset_update just sudo, remove concept of asset_sudo. 
- do a native transfer, taking in who they want to transfer to at a certain time.

pallets/automation-price/src/lib.rs
justinzhou93: it seems like yes. e.g. https://github.com/paritytech/substrate/blob/master/bin/node-template/pallets/template/src/lib.rs#L84. there's a read no matter what, even if not found

pallets/automation-price/Cargo.toml
```diff
@@ -0,0 +1,75 @@
...
+	"pallet-xcm/std",
+	"polkadot-parachain/std",
+	"scale-info/std",
+	"sp-runtime/std",
+	"sp-std/std",
+	"xcm/std",
+	"xcm-builder/std",
+	"xcm-executor/std",
+]
+dev-queue = []
```
rhuttman, line 75: do you need this?

rhuttman (approved):

pallets/automation-price/Cargo.toml
justinzhou93: nope, removing haha

PR #227, opened 2022-08-10
Move Tasks to AccountTasks

rhuttman:
This PR moves all tasks into a double map. This is from `Tasks<T>` to `AccountTasks<T>`. The keys for `AccountTasks<T>` are `account_id` then `task_id`. For those not familiar with double maps you are able to query either based on both keys or just the first key. That means we either need just the `account_id` or both the `account_id` and `task_id` to look up a task. This means that whenever we used to pass around the `task_id` we now need to pass around `(account_id, task_id)`.  This lead to the other items to storage to need be changed, along with a good amount of code. I did my best to make this a simple switch and did not take it upon myself to improve the code. 

I ask that you pay special attention to the migration and it's rather large test to ensure it was not correctly. You will notice I have also commented out, but kept, the previous migrations and add descriptions for what they did.


pallets/automation-time/src/lib.rs
```diff
@@ -665,39 +698,37 @@ pub mod pallet {
...
 		/// * `TaskDoesNotExist`: The task does not exist.
 		#[pallet::weight(<T as Config>::WeightInfo::cancel_scheduled_task_full())]
-		pub fn cancel_task(origin: OriginFor<T>, task_id: T::Hash) -> DispatchResult {
+		pub fn cancel_task(origin: OriginFor<T>, task_id: TaskId<T>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			match Self::get_task(task_id) {
-				None => Err(Error::<T>::TaskDoesNotExist)?,
-				Some(task) => {
-					if who != task.owner_id {
```
rhuttman, line 677: Due to the double map the user would just not find a task unless they own it

pallets/automation-time/src/lib.rs
```diff
@@ -665,39 +698,37 @@ pub mod pallet {
...
+		/// * `owner_id`: The owner of the task.
 		/// * `task_id`: The id of the task.
 		///
 		/// # Errors
 		/// * `TaskDoesNotExist`: The task does not exist.
 		#[pallet::weight(<T as Config>::WeightInfo::force_cancel_scheduled_task_full())]
-		pub fn force_cancel_task(origin: OriginFor<T>, task_id: T::Hash) -> DispatchResult {
+		pub fn force_cancel_task(
+			origin: OriginFor<T>,
+			owner_id: T::AccountId,
```
rhuttman, line 724: due to the double map we need the ownerId for the lookup

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,179 @@
...
+					}
+				})
+				.collect::<Vec<_>>();
+		MissedQueue2::<T>::put(new_missed_tasks);
+
+		// Set new storage version and return weight
+		StorageVersion::new(3).put::<Pallet<T>>();
+
+		// TODO set this as something
 		T::DbWeight::get().reads_writes(4, 4)
```
justinzhou93: ik you mentioned this already, but just remember to update this!

justinzhou93 (approved):
lgtm for the most part, prob worth testing the weight of the migration

pallets/automation-time/src/lib.rs
```diff
@@ -970,54 +1007,62 @@ pub mod pallet {
...
-							},
-						};
-						Self::decrement_task_and_remove_if_complete(*task_id, task);
-						task_action_weight
-							.saturating_add(T::DbWeight::get().writes(1 as Weight))
-							.saturating_add(T::DbWeight::get().reads(1 as Weight))
-					},
-				};
+				let action_weight =
+					match AccountTasks::<T>::get(account_task_id.0.clone(), account_task_id.1) {
```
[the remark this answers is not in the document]
rhuttman: Ya the diff got confused a few times

pallets/automation-time/src/lib.rs
```diff
@@ -155,11 +157,30 @@ pub mod pallet {
...
+			MissedTask2::<T> { owner_id, task_id, execution_time }
+		}
+	}
+
 	/// The struct that stores all information needed for a task.
 	#[derive(Debug, Eq, Encode, Decode, TypeInfo)]
 	#[scale_info(skip_type_params(T))]
 	pub struct Task<T: Config> {
-		owner_id: AccountOf<T>,
+		pub owner_id: AccountOf<T>,
```
nvengal, line 183: Does `owner_id` still make sense as a part of `Task` when the storage structure is:
```
owner_id:
  task_id:
    task: {
      owner_id
      ...
    }
```
Maybe this is outside the scope of this change, but we should atleast consider removing the duplicate owner_id in the future.

pallets/automation-time/src/lib.rs
```diff
@@ -406,54 +439,54 @@ pub mod pallet {
...
+			task_id: TaskId<T>,
 		},
 		/// Notify event for the task.
 		Notify {
 			message: Vec<u8>,
 		},
 		/// A Task was not found.
 		TaskNotFound {
-			task_id: T::Hash,
+			task_id: TaskId<T>,
```
nvengal, line 455: Should we be including the `account_id` as part of all the success and failure events since tasks can no longer be identified with just their task id?

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,212 @@
...
+			let account_task_ids: BoundedVec<AccountTaskId<T>, T::MaxTasksPerSlot> =
+				new_task_ids.try_into().unwrap();
+			ScheduledTasks2::<T>::insert(time, account_task_ids);
+			times_migrated += 1;
+		});
+
+		// Move all tasks from TaskQueue to TaskQueue2
+		let old_task_queue_prefix: &[u8] = b"TaskQueue";
+		let new_task_ids =
+			get_storage_value::<Vec<TaskId<T>>>(pallet_prefix, old_task_queue_prefix, &[])
```
nvengal, line 141: Looks like using `take_storage_value` might allow you to avoid calling `kill()` later. (I haven't verified that they're equivalent, but seem to be based on the docs)

The advantage I see here would be that you could delete all of the old storage types from the `lib.rs` file. Deleting the old storage types would remove dead code and prove that you didn't miss any code paths that are accidentally using the old storage still.

The extreme version of this would be to continue using the old (v1) storage prefixes and just modifying them in place, but that might be a bridge too far. It would reduce the size of this change set but it might make things more complex than necessary.

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,212 @@
...
+// 		LastTimeSlot::<T>::put((current_hour_slot, current_hour_slot));
+
+// 		log::info!(target: "automation-time", "Completed automation-time migration to v2");
+// 		StorageVersion::new(2).put::<Pallet<T>>();
+// 		T::DbWeight::get().reads_writes(4, 4)
+// 	}
+// }
+
+// Use a double map for tasks (accountId, taskId)
+pub mod v3 {
```
nvengal, line 80: This is a pretty hefty migration. The unit tests that you wrote are a good start, but I wonder if this might be a good opportunity to try out https://docs.substrate.io/reference/command-line-tools/try-runtime/ to test and demonstrate that this migration works as expected against our prod chain state.

pallets/automation-time/src/lib.rs
```diff
@@ -155,11 +157,30 @@ pub mod pallet {
 		}
 	}
 
+	/// The struct that stores data for a missed task.
+	#[derive(Debug, Eq, PartialEq, Encode, Decode, TypeInfo)]
+	#[scale_info(skip_type_params(T))]
+	pub struct MissedTask2<T: Config> {
```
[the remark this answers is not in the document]
rhuttman: It makes sense to me but happy to add the `V` to it

pallets/automation-time/src/lib.rs
```diff
@@ -970,54 +1007,62 @@ pub mod pallet {
...
 		/// Returns a vec with the tasks that were not run and the remaining weight.
 		pub fn run_tasks(
-			mut task_ids: Vec<T::Hash>,
+			mut account_task_ids: Vec<AccountTaskId<T>>,
 			mut weight_left: Weight,
-		) -> (Vec<T::Hash>, Weight) {
+		) -> (Vec<AccountTaskId<T>>, Weight) {
 			let mut consumed_task_index: usize = 0;
-			for task_id in task_ids.iter() {
+			for account_task_id in account_task_ids.iter() {
```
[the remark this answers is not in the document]
rhuttman: This is used in a lot of other places. I am fine removing the `account_task_id` type all together, but I think it needs to be consistent. This makes more sense to me, but if others find it confusing I am happy to remove the type and always spell it out

pallets/automation-time/src/lib.rs
```diff
@@ -87,6 +87,8 @@ pub mod pallet {
 		<<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
 	type UnixTime = u64;
 	type Seconds = u64;
+	pub type TaskId<T> = <T as frame_system::Config>::Hash;
+	pub type AccountTaskId<T> = (<T as frame_system::Config>::AccountId, TaskId<T>);
```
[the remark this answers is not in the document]
rhuttman: I would rather be explicit when creating a type.

pallets/automation-time/src/lib.rs
rhuttman: it does not make sense anymore, but is outside the scope. There is also no reason to keep `provided_id` in the struct that I can think of. I figured we will need to change the task struct soon anyway so can bundle the change then

pallets/automation-time/src/migrations.rs
rhuttman: ya I had been debating just removing the storage in this PR. For some reason I was hesitant to do it but I am fine making that change. 

I know I can use the same storage and just mutate the contents but I think new storage is safer

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
nvengal: I appreciated your use of account_task_id in the other places. I thought having a named type to spell it out was really good. Just right here, I think you'll benefit from destructuring the tuple so that in the rest of this method you can explicitly refer to account_id and task_id rather than account_task_id.0 and account_task_id.1

pallets/automation-time/src/lib.rs
rhuttman: hmmm I should probably change all the events who send `task_id` to also send `who`

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
nvengal: Not sure if this will clear up confusion, but this the code I was playing with to make sure that my suggestion was legal rust:
```
type T = (String, u64);
fn main() {
    let x: T = (String::from("a"), 1);
    let y: T = (String::from("b"), 2);
    let v = vec![x, y];
    for (s, n) in v.iter() {
        println!("{s:} {n:}");
    }
}
```

pallets/automation-time/src/lib.rs
```diff
@@ -155,11 +157,30 @@ pub mod pallet {
...
+	#[derive(Debug, Eq, PartialEq, Encode, Decode, TypeInfo)]
+	#[scale_info(skip_type_params(T))]
+	pub struct MissedTask2<T: Config> {
+		owner_id: T::AccountId,
+		task_id: TaskId<T>,
+		execution_time: UnixTime,
+	}
+
+	impl<T: Config> MissedTask2<T> {
+		pub fn create_missed_task(
```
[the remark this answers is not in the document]
nvengal: I prefer making the attributes public since it reduces the boilerplate code in the `lib.rs` file required for testing. But happy to defer to you guys if this is conscious design decision.

pallets/automation-time/src/lib.rs
nvengal: Works for me!

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
rhuttman: I just worry that if all places where we need to refer to different parts of the tuple we destruct it and all others we use the type that it could be confusing on where that type is actually relevant. I can make the change though

pallets/automation-time/src/migrations.rs
nvengal: > I know I can use the same storage and just mutate the contents but I think new storage is safer

Fine by me. I didn't actually know that you could migrate the storage in place. Learned that reading your pr, so thanks!

pallets/automation-time/src/migrations.rs
nvengal: I've never used try-runtime so I have no idea what I just asked you to do and maybe it's wildly more work than it's worth just figured it's something that we should explore as we deal with more complex state migrations.

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
rhuttman: I will do this in a follow on PR. I want to handle the MissedTask and Task struct at the same time

pallets/automation-time/src/migrations.rs
rhuttman: I have always wanted us to figure this out. Maybe this is the right excuse to spend on the time on it

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
nvengal: In my mind that's what the compiler is for. If we define our code with rigorous types then we can default to writing very readable code and the compiler can catch where the developer is misunderstanding the underlying type.

pallets/automation-time/src/migrations.rs
rhuttman: I am going to circle back to this after I test try-runtime. I want to make sure I don't need them for a two part migration

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,327 @@
...
+						} else {
+							log::debug!(target: "automation-time", "Unable to get task with id {:?}", task_id);
+							None
+						}
+					})
+					.collect::<Vec<_>>();
+
+				let account_task_ids: BoundedVec<AccountTaskId<T>, T::MaxTasksPerSlot> =
+					new_task_ids.try_into().unwrap();
+				ScheduledTasksV2::<T>::insert(time, account_task_ids);
```
justinzhou93, line 190: hm what's the likelihood that it's getting reinserted into the new map just as the scheduled time is about to pass? prob good to make sure no tasks are gonna run soon when you run migration, but no code change here

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,327 @@
...
+			storage_key_iter::<u64, BoundedVec<T::Hash, T::MaxTasksPerSlot>, Twox64Concat>(
+				pallet_prefix,
+				old_scheduled_prefix,
+			)
+			.drain()
+			.for_each(|(time, task_ids)| {
+				let new_task_ids = task_ids
+					.into_iter()
+					.filter_map(|task_id| {
+						if let Some(task) = old_tasks.get(&task_id) {
```
justinzhou93, line 179: wait how does this work? isn't old_tasks an arry?

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,327 @@
...
+			let new_task_ids =
+				take_storage_value::<Vec<TaskId<T>>>(pallet_prefix, old_task_queue_prefix, &[])
+					.unwrap_or(vec![])
+					.into_iter()
+					.filter_map(|task_id| {
+						if let Some(task) = old_tasks.get(&task_id) {
+							Some((task.owner_id.clone(), task_id))
+						} else {
+							log::debug!(target: "automation-time", "Unable to get task with id {:?}", task_id);
+							None
```
justinzhou93, line 205: i'm assuming filter removes None

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,327 @@
...
+				} else {
+					log::debug!(target: "automation-time", "Unable to get task with id {:?}", missed_task.task_id);
+					None
+				}
+			})
+			.collect::<Vec<_>>();
+			MissedQueueV2::<T>::put(new_missed_tasks);
+
+			// Set new storage version and return weight
+			StorageVersion::new(3).put::<Pallet<T>>();
```
justinzhou93, line 236: thought you were tossing the idea of storage version?

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,327 @@
...
+
+			// For each task there is
+			// 1 read to get it into memory from Tasks
+			// 1 write to remove it from Tasks
+			// 1 write to add it to AccountTasks
+			let weight = T::DbWeight::get().reads_writes(task_migrated + 1, task_migrated * 2);
```
justinzhou93, line 242: let mut necessary? or is it better to just redefine the variable each time?

pallets/automation-time/src/migrations.rs
```diff
@@ -1,72 +1,327 @@
...
+			// 1 read to get it into memory
+			// 1 write to remove it from the old task queue
+			// 1 write to add it to the new task queue
+			let weight = weight + T::DbWeight::get().reads_writes(1, 2);
+
+			// For the new storage version
+			let weight = weight + T::DbWeight::get().writes(1);
+
+			// Adding a buffer for the rest of the code
+			weight + 100_000_000
```
justinzhou93, line 267: 1 write's worth of buffer sufficient?

pallets/automation-time/src/weights.rs
```diff
@@ -230,4 +230,18 @@ impl<T: frame_system::Config> WeightInfo for AutomationWeight<T> {
...
+	// Storage: AutomationTime AccountTasks (r:0 w:6)
+	// Storage: AutomationTime TaskQueueV2 (r:0 w:1)
+	// Storage: unknown [0x85227ce1b17dca8228bad629dedea8b34e7b9012096b41c4eb3aaf947f6ea429] (r:0 w:1)
+	// Storage: AutomationTime ScheduledTasksV2 (r:0 w:2)
+	// Storage: AutomationTime MissedQueueV2 (r:0 w:1)
+	// fn migration_v3() -> Weight {
+	// 	(95_000_000 as Weight)
+	// 		.saturating_add(T::DbWeight::get().reads(12 as Weight))
+	// 		.saturating_add(T::DbWeight::get().writes(21 as Weight))
+	// }
```
justinzhou93, line 246: meant to leave the comment here? or mean to remove?

justinzhou93 (approved):
non blocking questions

pallets/automation-time/src/migrations.rs
```diff
@@ -260,10 +283,43 @@ pub mod v3 {
...
 
 			// Number of tasks in missed queue should not have changed
+			let missed_queue_prefix: &[u8] = b"MissedQueueV2";
+			let post_missed_queue_count =
+				get_storage_value::<Vec<MissedTaskV2<T>>>(pallet_prefix, missed_queue_prefix, &[])
+					.unwrap_or(vec![])
+					.len() as u32;
+			let pre_missed_queue_count =
+				Self::get_temp_storage::<u32>("pre_missed_queue_count").unwrap();
+			assert_eq!(post_missed_queue_count, pre_missed_queue_count);
```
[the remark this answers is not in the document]
rhuttman: they only run during try-runtime. and I have already tested that they work

pallets/automation-time/src/migrations.rs
rhuttman: why is that?  why would the code not handle it? the worst case a new time starts and we dont run the on_init for it. In this case its the same as a block from the previous time bleeding over. Which the code can handle

pallets/automation-time/src/migrations.rs
rhuttman: it it does

pallets/automation-time/src/migrations.rs
rhuttman: whoops

pallets/automation-time/src/migrations.rs
rhuttman: it's a btree

pallets/automation-time/src/migrations.rs
justinzhou93: ahh right, it's hours 🤦 , nvm you right haha

pallets/automation-time/src/migrations.rs
rhuttman: I like the redefine for this sort of thing

pallets/automation-time/src/migrations.rs
rhuttman: I think so

pallets/automation-time/src/weights.rs
rhuttman: mean to leave. It proves the weights I added to the migration

pallets/automation-time/src/migrations.rs
justinzhou93: just for my information, what's the diff? i.e. when should i redefine and when not?

rhuttman: it's more a style thing. I don't know a hard and fast rule yet

justinzhou93 (approved):

justinzhou93 (approved):

PR #228, opened 2022-08-11
Fix XCMP fee RPC error handling

atomaka:
* Pass the correct message through the runtime instead of printing a blatantly wrong message
* Propagate chain/currency combo error


runtime/turing/src/lib.rs
```diff
@@ -1241,7 +1242,14 @@ impl_runtime_apis! {
...
-				).map_err(|_| "cannot get xcmp fee")?.0;
+				).map_err(|e| {
+					match e {
+						DispatchError::Module(ModuleError{
+							message: Some("CurrencyChainComboNotFound"), ..
+						})=> "currency chain combination not supported",
+						_ => "cannot get xcmp fee"
+					}
+				})?.0;
 				let inclusion_fee = TransactionPayment::query_fee_details(extrinsic, len)
```
atomaka, line 1253: currency/chain combo error is important enough to pass through. Are there others? Is there a better way to do this?

runtime/turing/src/lib.rs
```diff
@@ -1233,15 +1234,22 @@ impl_runtime_apis! {
 		fn fees(encoded_xt: Bytes) -> Result<Balance, Vec<u8>> {
 			let extrinsic: <Block as BlockT>::Extrinsic = Decode::decode(&mut &*encoded_xt).unwrap();
 			if let Call::AutomationTime(pallet_automation_time::Call::schedule_xcmp_task{
-				provided_id, execution_times, para_id, currency_id, encoded_call, encoded_call_weight
+				execution_times, para_id, currency_id, encoded_call_weight, ..
 			}) = extrinsic.clone().function {
 				let len = encoded_xt.len() as u32;
 
 				let xcm_fee = XcmpHandler::calculate_xcm_fee_and_weight(
```
lreesby, line 1241: this fee also needs to be multiplied by `execution_times.len()`.

runtime/turing/src/lib.rs
lreesby: these are the other two:
```
		/// Either the weight or fee per second is too large.
		FeeOverflow,
		/// Either the instruction weight or the transact weight is too large.
		WeightOverflow,

runtime/turing/src/lib.rs
atomaka: I have another branch for resolving this. I'd prefer to release as separate PRs, but will leave that up to reviewers.

runtime/turing/src/lib.rs
atomaka: Updated to forward all module errors through.

rhuttman (approved):

lreesby (approved):
lgtm

PR #231, opened 2022-08-16
Adding try-runtime to Turing

rhuttman:


runtime/turing/src/lib.rs
```diff
@@ -1342,6 +1342,18 @@ impl_runtime_apis! {
...
+		fn on_runtime_upgrade() -> (Weight, Weight) {
+			log::info!("try-runtime::on_runtime_upgrade parachain-template.");
+			let weight = Executive::try_runtime_upgrade().unwrap();
+			(weight, RuntimeBlockWeights::get().max_block)
+		}
+
+		fn execute_block_no_check(block: Block) -> Weight {
+			Executive::execute_block_no_check(block)
+		}
+	}
```
justinzhou93, line 1356: not putting in for for neumann is fine, then i'm guessing? just keeping neumann around for splitting into OAK, right?

runtime/turing/src/lib.rs
```diff
@@ -1342,6 +1342,18 @@ impl_runtime_apis! {
 		}
 	}
 
+	#[cfg(feature = "try-runtime")]
+	impl frame_try_runtime::TryRuntime<Block> for Runtime {
+		fn on_runtime_upgrade() -> (Weight, Weight) {
+			log::info!("try-runtime::on_runtime_upgrade parachain-template.");
+			let weight = Executive::try_runtime_upgrade().unwrap();
```
justinzhou93, line 1349: wait docs say Executive::try_runtime_upgrade() should only be used for testing purposes. do we have a separate implementation for Prod?

node/src/command.rs
```diff
@@ -327,6 +327,26 @@ pub fn run() -> Result<()> {
...
+		#[cfg(feature = "try-runtime")]
+		Some(Subcommand::TryRuntime(cmd)) => {
+			let runner = cli.create_runner(cmd)?;
+			let chain_spec = &runner.config().chain_spec;
+			with_runtime_or_err!(chain_spec, {
+				{
+					// grab the task manager.
+					let registry =
+						&runner.config().prometheus_config.as_ref().map(|cfg| &cfg.registry);
+					let task_manager = sc_service::TaskManager::new(
```
justinzhou93, line 339: idk enough about this, but TaskManager seems to be for running async tasks. migrations wouldn't be running async, right? cause we'd probably want to run certain migrations synchronously (some less so, but for some, we'd probably want run migrations first before we start accepting extrinsics.)

runtime/turing/src/lib.rs
justinzhou93: i'm also confused what this does. it doesn't seem to be taking any specific migration. does it auto-detect migrations and run them even thought they're sitting in some folder and not mentioned here?

runtime/turing/src/lib.rs
```diff
@@ -1342,6 +1342,18 @@ impl_runtime_apis! {
...
+			(weight, RuntimeBlockWeights::get().max_block)
+		}
+
+		fn execute_block_no_check(block: Block) -> Weight {
```
justinzhou93, line 1353: curious difference between this function and on_runtime_upgrade. on_runtime_upgrade doesn't seem to be doing any checks either i think?

pallets/automation-time/src/lib.rs
```diff
@@ -469,17 +469,6 @@ pub mod pallet {
 				T::MaxWeightPercentage::get().mul_floor(T::MaxBlockWeight::get());
 			Self::trigger_tasks(max_weight)
 		}
-
-		fn on_runtime_upgrade() -> Weight {
-			let on_chain_storage_version = StorageVersion::get::<Pallet<T>>();
-			info!("on chain storage version, {:?}", on_chain_storage_version);
```
justinzhou93, line 475: we don't need on chain storage version?

runtime/turing/src/lib.rs
```diff
@@ -1342,6 +1342,18 @@ impl_runtime_apis! {
 		}
 	}
 
+	#[cfg(feature = "try-runtime")]
+	impl frame_try_runtime::TryRuntime<Block> for Runtime {
+		fn on_runtime_upgrade() -> (Weight, Weight) {
+			log::info!("try-runtime::on_runtime_upgrade parachain-template.");
+			let weight = Executive::try_runtime_upgrade().unwrap();
+			(weight, RuntimeBlockWeights::get().max_block)
```
justinzhou93, line 1350: why return the weight and max block size separately and not do the math?

justinzhou93 (commented):
lots of questions. i'm confused on a couple of things haha

runtime/turing/src/lib.rs
rhuttman: its useful for local testing, but since there is no live chain there is no need for try-runtime

runtime/turing/src/lib.rs
rhuttman: woops missed a log there. Try-runtime should never ever ever be run in prod as it takes extra cpu. You just compile w/ the flag on for testing your runtimes. It will attempt to perform a runtime upgrade if applicable. These can include migrations if you code says so.

rhuttman: this will make more sense when i add the new migration in the Double map pr

node/src/command.rs
rhuttman: this is for running the node w/ try-runtime

pallets/automation-time/src/lib.rs
rhuttman: We will be doing a different pattern for upgrades.  I will add in needed code to the double map PR

runtime/turing/src/lib.rs
rhuttman: Who says we only care about how mich is left

justinzhou93 (approved):

Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: try-runtime-turing
+try-runtime-turing:
+	RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace cargo run --release --features=turing-node,try-runtime try-runtime --chain turing-dev on-runtime-upgrade live --uri=wss://rpc.turing-staging.oak.tech:443
```
atomaka, line 3: `try-runtime-turing-staging` since it's running against staging, maybe?

node/Cargo.toml
```diff
@@ -38,6 +39,7 @@ codec = { package = "parity-scale-codec", version = "3.0.0" }
 serde = { version = "1.0.137", features = ["derive"] }
 serde_json = "1.0.68"
 hex-literal = "0.3.1"
+tracing-core = "=0.1.26"
```
atomaka, line 42: Do you think we need a comment here to remember to clean it up in a future release?

Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: try-runtime-turing
```
atomaka, line 1: Seems unlikely we'd ever have a file named `try-runtime-turing`, but probably good practice.

Also not sold on adding a `Makefile`.

atomaka (approved):

PR #236, opened 2022-08-17
ENG-478 Transact xcm functions

lreesby:


pallets/xcmp-handler/src/lib.rs
```diff
@@ -311,5 +343,70 @@ pub mod pallet {
...
+
+		/// Transact XCM instructions on local chain
+		///
+		pub fn transact_in_local_chain(
+			internal_instructions: xcm::v2::Xcm<<T as pallet::Config>::Call>,
+		) -> Result<(), DispatchError> {
+			let local_sovereign_account =
+				MultiLocation::new(1, X1(Parachain(T::SelfParaId::get().into())));
+			let weight = T::Weigher::weight(&mut internal_instructions.clone().into())
+				.map_or(Weight::max_value(), |w| T::BaseXcmWeight::get().saturating_add(w));
```
rhuttman, line 355: how does this base weight work? do we need it in the weight for the target chain instructions?

rhuttman (commented):
just 1 Q

pallets/xcmp-handler/src/lib.rs
lreesby: It is set here: https://github.com/OAK-Foundation/OAK-blockchain/blob/69b140aa21903f6f2a887a164a9e72899874c5ee/runtime/turing/src/xcm_config.rs#L404

lreesby: removed after chatting.

rhuttman (approved):

PR #239, opened 2022-08-18
Adding some helper scripts

rhuttman:


scripts/run-pallet-benchmarks.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+read -p 'Pallet name: ' pallet
+
+pallet_name=$(echo "$pallet" | tr '-' '_')
```
atomaka, line 5:
```suggestion
PS3="Pallet name: "
select pallet_name in $(ls pallets | tr '-' '_'); do break; done 
 ```

If you're feeling fancy.

File: scripts/run-pallet-benchmarks.sh [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+
+cargo run \
+--release \
```
atomaka, line 8: Generally see commands continued on the next line indented. ie.

```
cargo run \
  --release
```

atomaka (approved):
Minor comments; nothing blocking.

scripts/run-pallet-benchmarks.sh
rhuttman: I was wondering about this

PR #240, opened 2022-08-18
ENG-524 Add pallet-utility

lreesby:


antonkhvorov:
Amazing! 🚀

atomaka (changes requested):
I expected a Cargo.lock change with this.

lreesby:
> I expected a Cargo.lock change with this.

You're right.  I was making these changes through GHE while I ran benchmarks on another branch.  Will update.

justinzhou93 (approved):
seems like we're not really using it anywhere on the chain itself. more for users to use the batch extrinsic on it, right?

lreesby:
> seems like we're not really using it anywhere on the chain itself. more for users to use the batch extrinsic on it, right?

It's to allow users to make batch calls to our chain.

PR #241, opened 2022-08-18
First pass at  benchmark Github action

atomaka:
See https://github.com/OAK-Foundation/OAK-blockchain/pull/254 for a sample PR created by this request.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -71,6 +71,7 @@ jobs:
   generate-pull-request:
     name: Generate pull request
     needs: run-benchmarks
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
   # workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build-binary:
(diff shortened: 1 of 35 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,116 @@
...
+            --extrinsic '*' \
+            --repeat 20 \
+            --steps 50 \
+            --output ./${{ matrix.pallet }}-raw-weights.rs \
+            --template ./.maintain/frame-weight-template.hbs
+      - name: Upload raw weights
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.pallet }}
+          path: ${{ matrix.pallet }}-raw-weights.rs
```
nvengal, line 66: Any reason to have retention policy here as well?

nvengal (approved):

.github/workflows/benchmark.yml
atomaka: The default of 90 days seemed reasonable.  The only reason the `oak-collator` has one is that I don't consider it an artifact of the job. I just abuse artifacts to pass a specific version of the binary to a later job for use.

rhuttman (commented):
do you have an example output file for this?

atomaka:
> do you have an example output file for this?

@rhuttman The PR linked in the description (https://github.com/OAK-Foundation/OAK-blockchain/pull/254) is *basically* the output. The only difference is the I reverted the struct name change to match your PR. For completeness, here's one I just generated locally using this branch: https://gist.github.com/atomaka/484cef3331baa00ccbb470abe14527ea

rhuttman (approved):

PR #244, opened 2022-08-19
Removing old migration/storage

rhuttman:


justinzhou93 (approved):
lgtm! does https://github.com/OAK-Foundation/OAK-blockchain/blob/master/pallets/automation-time/src/weights.rs#L266 need to be removed too?

justinzhou93 (approved):

PR #255, opened 2022-08-25
Test all features in CI

nvengal:


.github/workflows/check.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       run: rustup show
     - name: Build
       run: |
-        cargo build --release --all --features all-nodes
+        cargo build --workspace --all-features
```
nvengal, line 40: `--all` is deprecated in favor of `--workspace` (https://doc.rust-lang.org/cargo/commands/cargo-build.html#package-selection)

nvengal: Removing `--release` significantly speeds up the build and test times. Since we're not publishing this binary, I don't think we need the `release` optimizations.
With `--release` (https://github.com/OAK-Foundation/OAK-blockchain/runs/8019045906?check_suite_focus=true)
- Build: 12m 4s
- Test: 4m 38s
Without `--release` (https://github.com/OAK-Foundation/OAK-blockchain/runs/8019517568?check_suite_focus=true)
- Build: 7m 57s
- Test: 1m 54s

nvengal: When we don't run tests for `--all-features` we can missing broken code in benchmarks. See: 2b87820da0bea99e84b72f17e5c58a8c79e9864e

rhuttman (approved):

justinzhou93 (approved):

PR #257, opened 2022-08-25
Use weight to determine automation slot capacity

nvengal:


pallets/automation-time/src/lib.rs
```diff
@@ -328,8 +272,7 @@ pub mod pallet {
 
 	#[pallet::storage]
 	#[pallet::getter(fn get_scheduled_tasks)]
-	pub type ScheduledTasksV2<T: Config> =
```
rhuttman, line 331: since you are removing this we need to remove `MaxTasksPerSlot`

pallets/automation-time/src/lib.rs
```diff
@@ -356,6 +299,7 @@ pub mod pallet {
 	pub type Shutdown<T: Config> = StorageValue<_, bool, ValueQuery>;
 
 	#[pallet::error]
+	#[derive(PartialEq)]
```
rhuttman, line 302: was this needed for a test?

pallets/automation-time/src/lib.rs
```diff
@@ -851,9 +795,11 @@ pub mod pallet {
 				}
 				MissedQueueV2::<T>::put(missed_queue);
 				// move current time slot to task queue or clear the task queue
-				if let Some(task_ids) = Self::get_scheduled_tasks(current_time_slot) {
+				if let Some(ScheduledTasks { tasks: task_ids, .. }) =
```
rhuttman, line 798: looks like I missed a rename here `account_task_ids`

pallets/automation-time/src/lib.rs
nvengal: yup, working on that now. good catch

pallets/automation-time/src/lib.rs
nvengal: yes, https://github.com/OAK-Foundation/OAK-blockchain/blob/71557422f5243f5acfd4c4b886425eec876a2558/pallets/automation-time/src/types.rs#L127-L131

pallets/automation-time/src/migrations.rs
```diff
@@ -320,3 +322,168 @@
...
+			log::info!(target: "automation-time", "Migrating automation-time v4");
+
+			// Move all tasks from ScheduledTasksV2 to ScheduledTasksV3
+			let mut task_count = 0u64;
+			let migrated_keys_count = storage_key_iter::<UnixTime, BoundedVec<AccountTaskId<T>, T::MaxTasksPerSlot>, Twox64Concat>(
+				PALLET_PREFIX,
+				OLD_STORAGE_PREFIX,
+			)
+			.drain()
+			.map(|(time, task_ids)| {
```
rhuttman, line 373: The map documentation says:

map() is conceptually similar to a for loop. However, as map() is lazy, it is best used when you're already working with other iterators. If you're doing some sort of looping for a side effect, it's considered more idiomatic to use for than map().

pallets/automation-time/src/migrations.rs
```diff
@@ -320,3 +322,168 @@
...
+				let weight = task_ids.clone()
+					.into_iter()
+					.filter_map(|(account_id, task_id)| {
```
rhuttman, line 376: could also use[ .fold](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold) for this

rhuttman (changes requested):
might be worth doing a benchmark for the migration to ensure you got the weight right

pallets/automation-time/src/migrations.rs
```diff
@@ -320,3 +322,168 @@
...
+			new_test_ext(0).execute_with(|| {
+				let task_id = TaskId::<Test>::default();
+				let account_id = AccountId32::new(ALICE);
+				let task = Task::<Test>::create_event_task(
+					account_id.clone(),
+					vec![0],
+					vec![0].try_into().unwrap(),
+					vec![0],
+				);
+				AccountTasks::<Test>::insert(account_id.clone(), task_id, task);
```
rhuttman, line 463: probably worth having more than one task and more than one scheduled time

runtime/turing/src/lib.rs
```diff
@@ -1016,11 +1017,13 @@ where
...
 impl pallet_automation_time::Config for Runtime {
 	type Event = Event;
-	type MaxTasksPerSlot = ConstU32<256>;
 	type MaxExecutionTimes = ConstU32<24>;
 	type MaxScheduleSeconds = MaxScheduleSeconds;
 	type MaxBlockWeight = MaxBlockWeight;
 	type MaxWeightPercentage = MaxWeightPercentage;
+	// Roughly .125% of parachain block weight per hour
+	// ≈ 500_000_000_000 (MaxBlockWeight) * 300 (Blocks/Hour) * .00125
+	type MaxWeightPerSlot = ConstU128<150_000_000_000>;
```
nvengal, line 1026: Roughly, 500 native transfer tasks per hour. Theoretically much lower than our actual capacity however I didn't want to couple this change with a significantly higher capacity (currently 256).

pallets/automation-time/src/benchmarking.rs
```diff
@@ -133,28 +130,33 @@ fn schedule_auto_compound_delegated_stake_tasks<T: Config>(
...
 			account_minimum,
 		);
+		let task_id = AutomationTime::<T>::schedule_task(&task, provided_id, vec![time]).unwrap();
 		<AccountTasks<T>>::insert(owner.clone(), task_id, task);
 		tasks.push((task_id, Pallet::<T>::get_account_task(owner.clone(), task_id).unwrap()));
 	}
 	tasks
 }
 
+fn increment_provided_id(mut provided_id: Vec<u8>) -> Vec<u8> {
```
nvengal, line 150: Left over change from when I was benchmarking with thousands of tasks. Leaving it in because we'll need something like this when we go back to realistic slot capacity.

pallets/automation-time/src/migrations.rs
```diff
@@ -320,3 +322,198 @@
...
+			// task_count reads to load tasks
+			// For each time (migrated_key_count) in scheduled tasks there is
+			// 1 read to get it into memory
+			// 1 write to remove it from the old scheduled tasks
+			// 1 write to add it to the new scheduled tasks
+			let weight = T::DbWeight::get()
+				.reads_writes(migrated_keys_count + task_count + 1, migrated_keys_count * 2);
+
+			// Adding a buffer for the rest of the code
+			weight + 10_000_000 + (migrated_keys_count * 15_000_000)
```
nvengal, line 402: Running benchmarks to prove these numbers

pallets/automation-time/src/types.rs
```diff
@@ -35,6 +37,21 @@ pub enum Action<AccountId, Balance, CurrencyId> {
 	},
 }
 
+impl<AccountId, Balance, CurrencyId> Action<AccountId, Balance, CurrencyId> {
+	pub fn execution_weight<T: Config>(&self) -> Weight {
```
rhuttman, line 41: out of curiosity did it not work to just use the `WeightInfo` trait?

rhuttman (commented):
once the weight is confirmed w/ benchmarks then lgtm

pallets/automation-time/src/types.rs
nvengal: It did not because of `Action::XCMP { .. } => T::DbWeight::get()` (T::DbWeight).

Also, didn't love that I had to import `crate::BalanceOf` from lib for https://github.com/OAK-Foundation/OAK-blockchain/pull/257/files#diff-62dc10adefef24a3ca74065c929abab1f72492b3bdd47293d2e7d92c4d340a26R221 but I'm not sure how to since the ScheduledTask struct doesn't have a Balance type this seemed like the most ergonomic way to express define the Task type expected by this method.

pallets/automation-time/src/migrations.rs
nvengal: @rhuttman mind double checking if I did something stupid in the benchmarks? I ran them overnight on my laptop and made very little progress. Doesn't seem to be an infinite loop because it is progressing just soooooo slowly.

nvengal: Oh. I think it's because notify tasks are so much cheaper to run. Current weight limits are set for ~500 transfers but that's ~15,000 notifies. Thoughts on how I should handle this? The diff between the cost of a notify task and the auto compound task are so large that it makes it difficult to set this limit.

pallets/automation-time/src/types.rs
rhuttman: you just need to add a generic to function to avoid this. Just like you did for the`execution_weight`

nvengal: I made the change you suggested here: https://github.com/OAK-Foundation/OAK-blockchain/pull/257/commits/64151e01af23168baaabae73f8fb783c7bcf3e98
I have some long term concern about how clunky and hard to read all of these generics are making the code. It's unclear to me what the value we're gaining by making these structs so generic. I expect the clunkiness and poor readability is only going to get worse as we write more complex code with this pattern.
But, that's a problem for a different pr!

pallets/automation-time/src/weights.rs
```diff
@@ -210,101 +218,121 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: AutomationTime AccountTasks (r:1 w:1)
 	// Storage: System Account (r:2 w:2)
 	/// The range of component `v` is `[0, 1]`.
 	fn run_tasks_many_found(v: u32, ) -> Weight {
-		(203_000 as Weight)
-			// Standard Error: 57_000
-			.saturating_add((39_164_000 as Weight).saturating_mul(v as Weight))
+		(0 as Weight)
+			// Standard Error: 100_000
+			.saturating_add((32_900_000 as Weight).saturating_mul(v as Weight))
```
nvengal, line 250: Flagging these zeroes for a second look. I expect that I screwed something up, but I'm not sure what.

pallets/automation-time/src/types.rs
```diff
@@ -185,3 +202,114 @@ impl<AccountId> TaskHashInput<AccountId> {
...
+		task: &Task<AccountId, Balance, T::CurrencyId, T::MaxExecutionTimes>,
+	) -> Result<&mut Self, Error<T>>
+	where
+		AccountId: Clone,
+	{
+		let weight = self
+			.weight
+			.checked_add(task.action.execution_weight::<T>() as u128)
+			.ok_or(Error::<T>::TimeSlotFull)?;
+		if self.tasks.len() as u32 >= T::MaxTasksPerSlot::get() ||
```
nvengal, line 230: Added MaxTasksPerSlot to avoid the risk of too many tasks being scheduled before we understand performance implications.

pallets/automation-time/src/types.rs
```diff
@@ -185,3 +202,114 @@ impl<AccountId> TaskHashInput<AccountId> {
...
+	pub tasks: Vec<(AccountId, TaskId)>,
+	pub weight: u128,
+}
+impl<A, B> Default for ScheduledTasks<A, B> {
+	fn default() -> Self {
+		Self { tasks: vec![], weight: 0 }
+	}
+}
+impl<AccountId, TaskId> ScheduledTasks<AccountId, TaskId> {
+	pub fn try_push<T: Config, Balance: AtLeast32BitUnsigned>(
```
rhuttman, line 218: you don't need `AtLeast32BitUnsigned`

pallets/automation-time/src/types.rs
```diff
@@ -185,3 +202,114 @@ impl<AccountId> TaskHashInput<AccountId> {
...
+	) -> Result<&mut Self, Error<T>>
+	where
+		AccountId: Clone,
+	{
+		let weight = self
+			.weight
+			.checked_add(task.action.execution_weight::<T>() as u128)
+			.ok_or(Error::<T>::TimeSlotFull)?;
+		if self.tasks.len() as u32 >= T::MaxTasksPerSlot::get() ||
+			weight > T::MaxWeightPerSlot::get()
```
rhuttman, line 231: isn't this condition met already in the check above?

pallets/automation-time/src/types.rs
rhuttman: probably worth a comment on why it is still needed

scripts/run-pallet-benchmarks.sh
```diff
@@ -17,4 +17,5 @@ cargo run \
     --extrinsic '*' \
     --repeat 20 \
     --steps 50 \
-    --output ./pallets/$pallet/src/raw-weights.rs
+    --output ./pallets/$pallet/src/raw-weights.rs \
+    --template ./.maintain/frame-weight-template.hbs
```
rhuttman, line 21: what is this for?

pallets/automation-time/src/weights.rs
rhuttman: Why update these weights? All weights should be updated before each release. I am not sure why you would need to do more than add your new weight fcns here

pallets/automation-time/src/benchmarking.rs
```diff
@@ -199,8 +201,14 @@ benchmarks! {
 		}
 
 		T::XcmpTransactor::setup_chain_currency_data(para_id.clone(), currency_id.clone())?;
-		let task_id: T::Hash = schedule_xcmp_tasks::<T>(caller.clone(), times.clone(), T::MaxTasksPerSlot::get() - 1);
-		let provided_id: Vec<u8> = vec![(T::MaxTasksPerSlot::get()/256).try_into().unwrap(), (T::MaxTasksPerSlot::get()%256).try_into().unwrap()];
+		let weight_per_task = ActionOf::<T>::XCMP {
```
rhuttman, line 204: I don't see this being used

pallets/automation-time/src/benchmarking.rs
```diff
@@ -229,19 +237,23 @@ benchmarks! {
...
 		}
 
-		let task_id: T::Hash = schedule_notify_tasks::<T>(caller.clone(), times.clone(), T::MaxTasksPerSlot::get() - 1);
-		let provided_id: Vec<u8> = vec![(T::MaxTasksPerSlot::get()/256).try_into().unwrap(), (T::MaxTasksPerSlot::get()%256).try_into().unwrap()];
+		let mut provided_id = schedule_notify_tasks::<T>(caller.clone(), times.clone(), T::MaxTasksPerSlot::get() - 1);
+		provided_id = increment_provided_id(provided_id);
 	}: schedule_native_transfer_task(RawOrigin::Signed(caller), provided_id, times, recipient, transfer_amount)
 
 	schedule_auto_compound_delegated_stake_task_full {
+		let max_tasks_per_slot: u32 = (
```
rhuttman, line 245: is this the one benchmark using `T::MaxWeightPerSlot` instead of `T::MaxTasksPerSlot` because it weighs so much it fails on  `T::MaxTasksPerSlot`?

rhuttman (changes requested):

pallets/automation-time/src/weights.rs
nvengal: I wanted to run all the benchmarks since I made code changes directly to the benchmarks and underlying implementations. Wanted to catch any errors I made earlier rather than later (like possibly this one). Happy to revert and only include my new benchmarks if it makes reviewing easier for you.
Regardless, I still expect that we will run actual benchmarks and replace these before the release.

pallets/automation-time/src/types.rs
nvengal:
```
.checked_add(task.action.execution_weight::<T>() as u128)
.ok_or(Error::<T>::TimeSlotFull)?;
```
only guards against overflow of the u128 type. This should never actually happen but since saturating could have unintended consequences on `>=` checks I opted to explicitly check the error
```
self.tasks.len() as u32 >= T::MaxTasksPerSlot::get()
```
Limits some of the smaller tasks but is not sufficient for heavier tasks like auto compounding (and in the future dynamic actions)

scripts/run-pallet-benchmarks.sh
nvengal: Matches the command used in github actions and makes the diff on the weights file way easier to read.
https://github.com/OAK-Foundation/OAK-blockchain/blob/19dc34639575ef8371b27520fcb9c96dc403df3b/.github/workflows/benchmark.yml#L61

pallets/automation-time/src/benchmarking.rs
nvengal: Yup, exactly.

pallets/automation-time/src/migrations.rs
nvengal: Resolved this by re-adding MaxTasksPerSlot as a hard limit for light weight tasks

pallets/automation-time/src/weights.rs
nvengal: Removed the extra weights

rhuttman (approved):

PR #262, opened 2022-08-29
Remove unused/outdated files

atomaka:
We have a lot of files we no longer use in the repository.

Todo

* `.maintain`: Some used, but most not used
* `.vscode`: Is anyone actively using / maintaining these configuration settings?
* `resources` vs `distribution`: Is distinction meaningful? Should we combine these directories?


nvengal:
Merging this in the current state to clear out existing prs. Todos remain todo.

nvengal (approved):

PR #265, opened 2022-08-30
add oak staging for genesis state/wasm

justinzhou93:


node/src/chain_spec/oak.rs
```diff
@@ -13,8 +13,11 @@ use primitives::{AccountId, AuraId, Balance};
 
 const TOKEN_SYMBOL: &str = "OAK";
 const SS_58_FORMAT: u32 = 51;
+const STAGING_SS_58_FORMAT: u32 = 52;
```
justinzhou93, line 16: i know we need to register our ss58 somewhere for this to work. but we need something different from 51 i think?

node/src/chain_spec/oak.rs
```diff
@@ -13,8 +13,11 @@ use primitives::{AccountId, AuraId, Balance};
...
 static RELAY_CHAIN: &str = "rococo-local";
+static STAGING_RELAY_CHAIN: &str = "rococo";
 const REGISTERED_PARA_ID: u32 = 2114;
+const REGISTERED_STAGING_PARA_ID: u32 = 2115;
```
justinzhou93, line 20: this paraID is open, but i think we need santiago to claim it before we can actually use

node/src/chain_spec/oak.rs
```diff
@@ -89,6 +92,81 @@ pub fn oak_development_config() -> ChainSpec {
...
+				ALLOC_TOKENS_TOTAL,
+				EXISTENTIAL_DEPOSIT,
+			);
+
+			testnet_genesis(
+				// initial collators.
+				vec![
+					(
+						// SS58 prefix substrate: 5EvKT4iWQ5gDnRhkS3d254RVCfaQJSyXHj5mA4kQTsCjWchW
+						hex!["7e4f4efde71551c83714fb3062724067e7bb6ebc3bf942813321f1583e187572"]
```
justinzhou93, line 124: all of these invulns, sudo, council, etc are all from turing staging. do we want different accounts?

justinzhou93 (commented):
generated genesis state/wasm off oak-dev, only to realize we can't use that 😐 . so couple of questions for @irsal :
1. ~are we going to be actually launching/maintaining an oak-staging on rococo or is this just for testing crowdloan?~
2. ~if we are going to maintaining an oak-staging, i had couple of questions on accounts/params~
3. ~need some ROC to test in Rococo - right now can't test contribute flow cause no ROC~
4. need the oak_staging_alloc.json to be updated with the correct values before genesis state is generated

node/src/chain_spec/oak.rs
irsal: nope, we can use 51

irsal: 51 is an address format. We'd like the address format to be the same across our chains: staging, turing, oak

justinzhou93: even if turing staging with ss58 51 already exists in rococo?

irsal: it's not. For example, mangatax uses the substrate default ss58 address. So do many of the parachains

justinzhou93: awesome, saves a bunch of work. will use the same 51 ss58!

justinzhou93:
image
image

node/src/chain_spec/oak.rs
irsal: We do; not blocking this pr. I will change those after I gathered all wallets

irsal (changes requested):
let's just do oak (no staging)

node/src/chain_spec/oak.rs
```diff
@@ -89,6 +95,77 @@ pub fn oak_development_config() -> ChainSpec {
...
+pub fn oak_staging() -> ChainSpec {
+	// Give your base currency a unit name and decimal places
+	let mut properties = sc_chain_spec::Properties::new();
+	properties.insert("tokenSymbol".into(), TOKEN_SYMBOL.into());
+	properties.insert("tokenDecimals".into(), TOKEN_DECIMALS.into());
+	properties.insert("ss58Format".into(), SS_58_FORMAT.into());
+
+	ChainSpec::from_genesis(
+		// Name
+		"Oak Staging",
```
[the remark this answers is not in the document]
justinzhou93: we want it to be called `OAK` for the staging instance too? this is the one on rococo

node/src/chain_spec/oak.rs
justinzhou93: cool, sounds good - just fyi, can't do genesis things for staging until this is in tho!

irsal: sudo should be `5C571x5GLRQwfA3aRtVcxZzD7JnzNb3JbtZEvJWfQozWE54K`

justinzhou93: yes, i believe it should be! double check me, but i'm pretty sure that's what it's set as! (it's the same as turing staging)

justinzhou93:
image
vesting looks good - timestamps are in the past, cause i just copy pasted from turing staging, but should match.

only thing looks like the timestamps aren't in the right order, but it does look like this is the case in turing as well.

distribution/oak_staging_alloc.json
```diff
@@ -0,0 +1,5 @@
+[
```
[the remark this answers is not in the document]
justinzhou93: done!

node/src/chain_spec/oak.rs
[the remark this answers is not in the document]
justinzhou93: chatted on slack - keeping oak-staging

irsal (approved):

PR #266, opened 2022-09-01
ENG-548: Add orml-asset-registry pallet

lreesby:


runtime/turing/src/xcm_config.rs
```diff
@@ -180,107 +180,13 @@ pub type Barrier = DenyThenTry<
...
-
-/// Assuming ~ $0.50 TUR price.
-pub fn tur_per_second() -> u128 {
-	let tur_equivalent = convert_to_token(CurrencyId::KSM, CurrencyId::Native, ksm_per_second());
-	// Assuming KSM ~ $130.00.
-	tur_equivalent * 260
-}
-
-parameter_types! {
-	pub UnitPerSecond: (AssetId, u128) = (
```
justinzhou93, line 196: i know this isn't quite done yet, but i was just curious about asset registry and poking around in orml. us removing this is because we can toss all this information into the asset registry storage? https://github.com/open-web3-stack/open-runtime-module-library/blob/master/asset-registry/src/lib.rs#L91

lreesby: Correct.  This information will be stored in the orml asset registry.  ie.  feePerSec, tokenConversionRate (KSM:Native, etc), and decimals.

node/src/chain_spec/turing.rs
```diff
@@ -160,6 +160,7 @@ fn testnet_genesis(
 		treasury: Default::default(),
 		valve: ValveConfig { start_with_valve_closed: false, closed_gates: pallet_gates_closed },
 		vesting: VestingConfig { vesting_schedule },
+		asset_registry: Default::default(),
```
rhuttman, line 163: this will need to change in order for https://github.com/OAK-Foundation/OAK-blockchain/pull/269 to work. Does not need to be in this PR though.

node/src/chain_spec/oak.rs
```diff
@@ -149,5 +149,6 @@ fn testnet_genesis(
 		treasury: Default::default(),
 		valve: ValveConfig { start_with_valve_closed: false, closed_gates: pallet_gates_closed },
 		vesting: VestingConfig { vesting_schedule },
+		asset_registry: Default::default(),
```
rhuttman, line 152: should we not add `Native` by default?

runtime/neumann/src/lib.rs
```diff
@@ -164,6 +158,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	state_version: 0,
 };
 
+pub const NATIVE_TOKEN_ID: TokenId = 0;
```
rhuttman, line 161: you need to use this in the gensis config to make this true right?

runtime/neumann/src/lib.rs
```diff
@@ -529,6 +479,30 @@ impl orml_currencies::Config for Runtime {
...
+}
+
+impl orml_asset_registry::Config for Runtime {
+	type Event = Event;
+	type CustomMetadata = CustomMetadata;
+	type AssetId = TokenId;
+	type AuthorityOrigin = AssetAuthority;
+	type AssetProcessor = orml_asset_registry::SequentialId<Runtime>;
+	type Balance = Balance;
+	type WeightInfo = ();
```
rhuttman, line 503: these weights are all 0..... We probably need to run the benchmarks ourselves

rhuttman (changes requested):
I still need to review the migration

runtime/neumann/src/lib.rs
lreesby: You can’t add assets in the genesis config.

runtime/turing/src/lib.rs
```diff
@@ -470,36 +472,11 @@ parameter_types! {
...
-			CurrencyId::KAR => 10 * CurrencyId::KAR.cent(),
-			CurrencyId::LKSM => 50 * CurrencyId::LKSM.millicent(),
-			CurrencyId::HKO => 50 * CurrencyId::HKO.cent(),
-			CurrencyId::SKSM => 50 * CurrencyId::SKSM.millicent(),
-			CurrencyId::PHA => CurrencyId::PHA.cent(),
-		}
-	};
-}
-
+// TODO: remove after asset registry migration
```
nvengal, line 475: Can this be moved to the migration file? It doesn't look like it depends on any state in the runtime.

runtime/turing/src/migrations.rs
```diff
@@ -0,0 +1,526 @@
...
+
+			log::info!(
+				target: "asset_registry",
+				"on_runtime_upgrade: New data inserted"
+			);
+
+			// Each asset + each asset location + updating last asset id
+			let total_rw = assets.len() as u32 * 2 + 1;
+			<Runtime as frame_system::Config>::DbWeight::get()
+				.reads_writes(total_rw as Weight, total_rw as Weight)
```
nvengal, line 300: Is `try_mutate` 1 read + 1 write? I expected this to be just `.writes`.

runtime/turing/src/migrations.rs
```diff
@@ -0,0 +1,526 @@
...
+
+	pub struct AssetRegistryMigration;
+	impl OnRuntimeUpgrade for AssetRegistryMigration {
+		fn on_runtime_upgrade() -> Weight {
+			log::info!(
+				target: "asset_registry",
+				"on_runtime_upgrade: Attempted to apply asset_registry migration"
+			);
+
+			let assets = vec![
```
nvengal, line 110: What do you think about moving this vec outside of `on_runtime_upgrade`? It would make this method easier to read and you could assert on `vec.len()` in the post_upgrade.

runtime/turing/src/migrations.rs
```diff
@@ -0,0 +1,526 @@
...
+			log::info!(
+				target: "orml_tokens",
+				"on_runtime_upgrade: Attempted to apply tokens"
+			);
+
+			// Migrate Tokens Accounts CurrencyId from Enum to u32
+			let mut tokens_accounts: Vec<(AccountId, TokenId, orml_tokens::AccountData<Balance>)> =
+				vec![];
+			Accounts::drain().for_each(|(account_id, currency_id, account_data)| {
+				tokens_accounts.push((account_id, currency_id as u32, account_data));
```
nvengal, line 372: Assuming that you verified that taking the `currency_id as u32` preserves the order and value expected by the rest of this code. I expect that rust enums are just 0 indexed ints under the hood, but I'm not sure.

runtime/turing/src/migrations.rs
```diff
@@ -0,0 +1,526 @@
...
+			});
+			tokens_accounts.iter().for_each(|(account_id, currency_id, account_data)| {
+				orml_tokens::Accounts::<Runtime>::insert(account_id, currency_id, account_data);
+			});
+			log::info!(
+				target: "orml_tokens",
+				"on_runtime_upgrade: Accounts data updated {:?}",
+				orml_tokens::Accounts::<Runtime>::iter().count(),
```
nvengal, line 380: Does this cause a full set of extra reads? I'm not sure how substrate does `iter().count()`.

runtime/neumann/src/xcm_config.rs
```diff
@@ -388,51 +354,30 @@ impl pallet_xcmp_handler::Config for Runtime {
...
-					// testchain
-					id if id == u32::from(parachains::testchain::ID) => Some(CurrencyId::UNIT),
-					_ => None,
-				}
-			},
-			// adapt for re-anchor canonical location: https://github.com/paritytech/polkadot/pull/4470
-			MultiLocation { parents: 0, interior: Here } => Some(CurrencyId::Native),
-			_ => None,
+			// adapt for re-anchor canonical location bug: https://github.com/paritytech/polkadot/pull/4470
+			MultiLocation { parents: 1, interior: X1(Parachain(para_id)) }
```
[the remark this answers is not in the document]
lreesby: `MultiLocation { parents: 0, interior: Here }` is what is expected and exists in the assetRegistry as our Native token. https://github.com/OAK-Foundation/OAK-blockchain/blob/laura/orml-asset-registry/runtime/turing/src/migrations.rs#L119

`MultiLocation { parents: 1, interior: X1(Parachain(para_id)) }` accounts for the location the wasn't properly being re-anchored.

runtime/neumann/src/lib.rs
```diff
@@ -529,6 +479,30 @@ impl orml_currencies::Config for Runtime {
...
+	#[cfg(feature = "runtime-benchmarks")]
+	fn successful_origin(_asset_id: &Option<u32>) -> Origin {
+		EnsureRoot::successful_origin()
+	}
+}
+
+impl orml_asset_registry::Config for Runtime {
+	type Event = Event;
+	type CustomMetadata = CustomMetadata;
+	type AssetId = TokenId;
```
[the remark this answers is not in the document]
lreesby: Users would have to check the assets in the `AssetRegistry` to know what asset they want to play with.  It's an extra read to the chain state to verify but I would argue that even if we had descriptive enums you would probably want to verify the multilocation you are playing with to be certain.  Ultimately, IMO, the enums or even a longer string won't ever be as sufficient as verifying the multilocation.  Especially when you're moving funds around.

justinzhou93 (commented):
remember to add to `ClosedCallFilter`!

rhuttman (approved):

runtime/turing/src/migrations.rs
```diff
@@ -0,0 +1,539 @@
...
+			match a {
+				0 => CurrencyId::Native,
+				1 => CurrencyId::KSM,
+				2 => CurrencyId::AUSD,
+				3 => CurrencyId::KAR,
+				4 => CurrencyId::LKSM,
+				5 => CurrencyId::HKO,
+				6 => CurrencyId::SKSM,
+				7 => CurrencyId::PHA,
+				8 => CurrencyId::UNIT,
```
nvengal, line 39: Is this a good opportunity to get rid of/ not migrate this?

nvengal (approved):

PR #270, opened 2022-09-06
Allow technical committee to use Valve 

rhuttman:


pallets/valve/src/benchmarking.rs
```diff
@@ -19,45 +19,81 @@
...
+	verify {
+		assert_last_event::<T>(Event::ScheduledTasksStopped.into())
+	}
 
 	start_scheduled_tasks {
+		let caller = get_caller_account_id::<T>();
 		T::AutomationTime::shutdown();
-	}: _(RawOrigin::Root)
+	}: _(RawOrigin::Signed(caller))
+	verify {
```
justinzhou93, line 96: there was general weight increase across the board in the weights. was it because of the verify? or was it really just because there's an extra if in the logic?

justinzhou93 (approved):
1 Q, but looks good

pallets/valve/src/benchmarking.rs
rhuttman: Verify does not add weight. It's just a test to the state after the benchmark is run

pallets/valve/src/benchmarking.rs
```diff
@@ -19,45 +19,81 @@
...
+	T::CallAccessFilter::sorted_members().pop().unwrap()
+}
+
 benchmarks! {
 	close_valve {
-
-	}: _(RawOrigin::Root)
+		let caller = get_caller_account_id::<T>();
+	}: _(RawOrigin::Signed(caller))
+	verify {
```
nvengal, line 37: 💯

pallets/valve/src/mock.rs
```diff
@@ -87,6 +90,21 @@ impl Contains<Call> for ClosedCallFilter {
...
 }
 
+pub struct TechCollective;
+impl SortedMembers<AccountId> for TechCollective {
+	fn contains(account_id: &AccountId) -> bool {
+		if account_id == &AccountId32::new(COLLECTIVE_MEMBER) {
+			true
+		} else {
+			false
+		}
```
nvengal, line 100:
```suggestion
		account_id == &AccountId32::new(COLLECTIVE_MEMBER)
```

nvengal (approved):

pallets/valve/src/mock.rs
rhuttman: oh duh

nvengal (approved):

PR #271, opened 2022-09-06
Adding call filter to dynamic dispatch

rhuttman:


justinzhou93 (approved):
lgtm - seems like it's going to be allowing or rejecting entire pallets. ever gonna be a scenario where we only wanna allow for certain extrinsics within a pallet to be able to go through?

rhuttman:
@justinzhou93 This will also work for that use case.

nvengal (approved):

PR #278, opened 2022-09-13
TUR and OAK runtime config changes

justinzhou93:


runtime/oak/src/lib.rs
```diff
@@ -673,11 +673,11 @@ parameter_types! {
...
-	pub const DefaultBlocksPerRound: u32 = 2 * HOURS;
+	pub const DefaultBlocksPerRound: u32 = 6 * HOURS;
 	/// Minimum stake required to become a collator
-	pub const MinCollatorStk: u128 = 400_000 * DOLLAR;
+	pub const MinCollatorStk: u128 = 800_000 * DOLLAR;
 	/// Minimum stake required to be reserved to be a candidate
-	pub const MinCandidateStk: u128 = 2_000_000 * DOLLAR;
+	pub const MinCandidateStk: u128 = 800_000 * DOLLAR;
 }
 impl pallet_parachain_staking::Config for Runtime {
```
irsal, line 682: @nvengal - please review all the parachain-staking changes. See Jira tix for requirements: https://oak-network.atlassian.net/browse/ENG-621

node/src/chain_spec/oak.rs
```diff
@@ -21,8 +21,10 @@ const TOKEN_SYMBOL: &str = "OAK";
 const SS_58_FORMAT: u32 = 51;
 static RELAY_CHAIN: &str = "rococo-local";
 static STAGING_RELAY_CHAIN: &str = "rococo";
+static LIVE_RELAY_CHAIN: &str = "polkadot";
 const REGISTERED_PARA_ID: u32 = 2114;
 const REGISTERED_STAGING_PARA_ID: u32 = 4103;
+const REGISTERED_LIVE_PARA_ID: u32 = 2090;
```
rhuttman, line 27: why not just use the `REGISTERED_PARA_ID`?

node/src/chain_spec/oak.rs
```diff
@@ -188,6 +190,105 @@ pub fn oak_staging() -> ChainSpec {
...
+		"OAK",
+		// ID
+		"oak",
+		ChainType::Live,
+		move || {
+			let allocation_json =
+				&include_bytes!("../../../distribution/oak_alloc.json")[..];
+			let initial_allocation: Vec<(AccountId, Balance)> =
+				serde_json::from_slice(allocation_json).unwrap();
+			const ALLOC_TOKENS_TOTAL: u128 = DOLLAR * 1_000_000_000;
```
rhuttman, line 211: the initial alloc is much lower right?

justinzhou93: yah, waiting for vesting and alloc to come through

runtime/turing/src/lib.rs
```diff
@@ -922,11 +922,11 @@ impl pallet_scheduler::Config for Runtime {
 parameter_types! {
 	pub const LaunchPeriod: BlockNumber = 5 * DAYS;
 	pub const VotingPeriod: BlockNumber = 5 * DAYS;
-	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
+	pub const FastTrackVotingPeriod: BlockNumber = 1 * HOURS;
 	pub const MinimumDeposit: Balance = 40 * DOLLAR;
-	pub const EnactmentPeriod: BlockNumber = 2 * DAYS;
-	pub const VoteLockingPeriod: BlockNumber = 7 * DAYS;
-	pub const CooloffPeriod: BlockNumber = 7 * DAYS;
+	pub const EnactmentPeriod: BlockNumber = 1 * DAYS;
```
irsal, line 927: @justinzhou93 - let's bump this back to 2 days per @lreesby 's feedback

runtime/oak/src/lib.rs
```diff
@@ -1059,7 +1045,7 @@ impl pallet_valve::Config for Runtime {
 	type WeightInfo = pallet_valve::weights::SubstrateWeight<Runtime>;
 	type ClosedCallFilter = ClosedCallFilter;
 	type AutomationTime = AutomationTime;
-	type AutomationPrice = AutomationPrice;
+	type AutomationPrice = AutomationTime;
```
nvengal, line 1048: This is weird. Why not just add a dummy implementation to `()`?

node/src/chain_spec/oak.rs
```diff
@@ -196,6 +197,115 @@ pub fn oak_staging() -> ChainSpec {
...
+							.unchecked_into(),
+					),
+					(
+						// SS58 prefix substrate: 5F4Dx9TD16awU7FeGD3Me5VDrEsL5MDPEcbNBvUgLQFawxyW
+						hex!["8456c30af083c2b2d767d6950e8ee654d4eff5e69cdd9f75db5bcbf2d7b0d801"]
+							.into(),
+						hex!["8456c30af083c2b2d767d6950e8ee654d4eff5e69cdd9f75db5bcbf2d7b0d801"]
+							.unchecked_into(),
+					),
+				],
```
nvengal, line 256: Any reason to not include all 4 four of the initial collators that we plan to launch with?
Also, double checking that these are new accounts (not used on other networks) and that someone has the secret seed stored. We will need to provide the seed to the collator nodes for the chain to start producing.

runtime/oak/src/lib.rs
nvengal: left some other comments but 👍 on parachain staking runtime config

runtime/oak/src/lib.rs
```diff
@@ -673,11 +672,11 @@ parameter_types! {
...
 	pub const DefaultParachainBondReservePercent: Percent = Percent::from_percent(30);
 	/// Blocks per round
-	pub const DefaultBlocksPerRound: u32 = 2 * HOURS;
+	pub const DefaultBlocksPerRound: u32 = 6 * HOURS;
 	/// Minimum stake required to become a collator
-	pub const MinCollatorStk: u128 = 400_000 * DOLLAR;
+	pub const MinCollatorStk: u128 = 800_000 * DOLLAR;
 	/// Minimum stake required to be reserved to be a candidate
-	pub const MinCandidateStk: u128 = 2_000_000 * DOLLAR;
+	pub const MinCandidateStk: u128 = 800_000 * DOLLAR;
```
irsal, line 679: Bumping this to 2MM

irsal: min collator bond is 2mm, so both 677 and 679

node/src/chain_spec/oak.rs
```diff
@@ -196,6 +197,115 @@ pub fn oak_staging() -> ChainSpec {
...
+			let vest_ending_time: u64 = 1764568800;
+			validate_vesting(
+				initial_vesting.clone(),
+				vested_tokens,
+				EXISTENTIAL_DEPOSIT,
+				vest_starting_time,
+				vest_ending_time,
+			);
+
+			testnet_genesis(
```
irsal, line 239: make sure to update this wihtt he correct collators

justinzhou93: where do i find the right collator addresses?

node/src/chain_spec/oak.rs
irsal: These are 4 new accounts, yes

runtime/oak/src/lib.rs
justinzhou93:
```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
      --> /Users/justinzhou/Documents/OAK/ENG/OAK-blockchain/runtime/oak/src/lib.rs:1043:1
       |
  1043 | impl pallet_valve::Shutdown for () {
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^--
       | |                               |
       | |                               this is not defined in the current crate because tuples are always foreign
       | impl doesn't use only types from inside the current crate
       |
       = note: define and implement a trait or new type instead
```

it didn't like impl for (), so just did impl on a new struct called DummyPallet. Lemme know if this works!

nvengal: You need to define the `impl Shutdown for ()` in the valve pallet. Probably makes sense to do so right in `valve/src/traits.rs` file as a default implementation.

The rule is that either the trait being implemented or the struct on which its being implemented needs to be defined within the crate doing the implementation. Since neither `Shutdown` or `()` are defined in the runtime it can't be implemented here.

justinzhou93: ahh i gotcha, thanks!

node/src/chain_spec/oak.rs
```diff
@@ -21,7 +21,8 @@ const TOKEN_SYMBOL: &str = "OAK";
 const SS_58_FORMAT: u32 = 51;
 static RELAY_CHAIN: &str = "rococo-local";
 static STAGING_RELAY_CHAIN: &str = "rococo";
-const REGISTERED_PARA_ID: u32 = 2114;
+static LIVE_RELAY_CHAIN: &str = "polkadot";
```
irsal, line 24: we confirmed this from their chainspec, right?

node/src/chain_spec/oak.rs
```diff
@@ -190,6 +191,129 @@ pub fn oak_staging() -> ChainSpec {
...
+				vec![
+					b"AutomationTime".to_vec(),
+					b"Balances".to_vec(),
+					b"Bounties".to_vec(),
+					b"Currencies".to_vec(),
+					b"Democracy".to_vec(),
+					b"ParachainStaking".to_vec(),
+					b"PolkadotXcm".to_vec(),
+					b"Treasury".to_vec(),
+					b"XTokens".to_vec(),
```
irsal, line 278: is this the right capitalization? `XTokens` vs `xTokens`?

pallets/valve/src/traits.rs
```diff
@@ -7,3 +7,11 @@ pub trait Shutdown {
 	/// Forwards restart message to implementer
 	fn restart();
 }
+
+impl Shutdown for () {
+	fn is_shutdown() -> bool {
```
irsal, line 12: what does this mean?

runtime/oak/src/lib.rs
```diff
@@ -465,7 +464,7 @@ parameter_types! {
 	pub TreasuryAccount: AccountId = TreasuryPalletId::get().into_account_truncating();
 	// Until we can codify how to handle forgien tokens that we collect in XCMP fees
 	// we will send the tokens to a special account to be dealt with.
```
irsal, line 466: add comment on the ss58 `51` wallet address please

node/src/chain_spec/oak.rs
justinzhou93: yah. naming comes from runtime files: https://github.com/OAK-Foundation/OAK-blockchain/blob/master/runtime/oak/src/lib.rs#L1083

node/src/chain_spec/oak.rs
justinzhou93: believe so! https://raw.githubusercontent.com/paritytech/polkadot/master/node/service/chain-specs/polkadot.json

pallets/valve/src/traits.rs
justinzhou93: this is because we ripped out automationPrice pallet. Valve expects automation pallet and automation time pallets to be passed into it so that it can shut down the pallet as needed. In this case, since we removed automation price, there's nothing to pass into Valve. Therefore, we have to pass a "mock" item into valve. What valve is expecting is an object with a trait `Shutdown`, as defined above. A [similar implementation](https://github.com/OAK-Foundation/OAK-blockchain/blob/master/pallets/automation-price/src/lib.rs#L880) was made in automation price. Here, rather than defining a mock automationPallet, we just used `()` and passed that into Valve instead to stub the missing pallet

node/src/chain_spec/oak.rs
justinzhou93: hmmm actually now that i look at the other chain specs in that file (kusama and rococo), our names don't quite match the names on their chain spec

rhuttman (approved):

justinzhou93 (commented):
double checking - no need to bump the node version, right? since only node files i changed are chain_spec files and command file (adding oak), both of which aren't going to be touched by existing turing chains

irsal (approved):
I triple checked the addresses. LGTM!

Thanks for doing this!

nvengal (approved):

PR #281, opened 2022-09-13
Removing unneeded Neumann chain specs

rhuttman:


node/src/command.rs
```diff
@@ -23,12 +23,6 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
 	Ok(match id {
 		#[cfg(feature = "neumann-node")]
 		"dev" => Box::new(chain_spec::neumann::development_config()),
-		#[cfg(feature = "neumann-node")]
-		"" | "local" => Box::new(chain_spec::neumann::local_testnet_config()),
```
nvengal, line 27: do you want to change the `""` string to be turing-dev or just "dev"?

rhuttman: do you mean nuemann-dev?

node/src/chain_spec/neumann.rs
```diff
@@ -111,196 +104,6 @@ pub fn development_config() -> ChainSpec {
...
-		// Extensions
-		Extensions {
-			relay_chain: NEUMANN_RELAY_CHAIN.into(), // You MUST set this to the correct network!
-			para_id: DEFAULT_PARA_ID,
-		},
-	)
-}
-
-pub fn neumann_latest() -> Result<DummyChainSpec, String> {
-	DummyChainSpec::from_json_bytes(&include_bytes!("../../res/neumann.json")[..])
```
nvengal, line 301: Can probably remove the neumann files from in here while you're at it- https://github.com/OAK-Foundation/OAK-blockchain/tree/master/node/res

node/src/command.rs
nvengal: this one.
`"dev" => Box::new(chain_spec::neumann::development_config()),`
doesn't look like there is a `neumann-dev`

nvengal: Sorry, that still came out confusing.
On L27 here you've deleted the default option when someone doesn't specify a chain (`"" | "local"`). I was questioning whether there should be a new default.

PR #282, opened 2022-09-15
Formalize scheduling variants

nvengal:
This pr adds the `Schedule` enum to `Tasks` to formalize the implicit scheduling variants.
Currently these are either `Fixed` or `Recurring`.
A follow up to this PR will add `Schedule` as an argument to the dynamic dispatch (and maybe xcmp) scheduling extrinsics.

### DIFF
--- a/pallets/automation-time/src/benchmarking.rs
+++ b/pallets/automation-time/src/benchmarking.rs
@@ -498,30 +498,30 @@ benchmarks! {
 		schedule_notify_tasks::<T>(caller.clone(), vec![new_time_slot], T::MaxTasksPerSlot::get());
 	}: { AutomationTime::<T>::shift_missed_tasks(last_time_slot, diff) }
 
-	migration_v5 {
+	migration_add_schedule_to_task {
 		let v in 1..100;
 
-		use migrations::*;
+		use migrations::add_schedule_to_task::{AccountTasks as OldAccountTasks, AddScheduleToTask, OldAction, OldTask};
 		use frame_support::traits::OnRuntimeUpgrade;
 
 		for i in 0..v {
 			let account_id: T::AccountId = account("Account", 0, i);
 			let task_id_1 = Pallet::<T>::generate_task_id(account_id.clone(), vec![0]);
-			let task_1 = v5::OldTask::<T> {
+			let task_1 = OldTask::<T> {
 				owner_id: account_id.clone(),
 				provided_id: vec![1],
 				execution_times: vec![10].try_into().unwrap(),
 				executions_left: 1,
-				action: v5::OldAction::<T>::AutoCompoundDelegatedStake {
+				action: OldAction::<T>::AutoCompoundDelegatedStake {
 					delegator: account_id.clone(),
 					collator: account_id.clone(),
 					account_minimum: 100u32.into(),
 					frequency: 11,
 				},
 			};
-			v5::AccountTasks::<T>::insert(account_id.clone(), task_id_1, task_1);
+			OldAccountTasks::<T>::insert(account_id.clone(), task_id_1, task_1);
 		}
-	}: { v5::MigrateToV5::<T>::on_runtime_upgrade() }
+	}: { AddScheduleToTask::<T>::on_runtime_upgrade() }
 	verify {
 		assert_eq!(v, crate::AccountTasks::<T>::iter().count() as u32);
 	}
--- a/pallets/automation-time/src/benchmarking.rs
+++ b/pallets/automation-time/src/benchmarking.rs
@@ -506,8 +506,8 @@ benchmarks! {
 
 		for i in 0..v {
 			let account_id: T::AccountId = account("Account", 0, i);
-			let task_id_1 = Pallet::<T>::generate_task_id(account_id.clone(), vec![0]);
-			let task_1 = OldTask::<T> {
+			let task_id = Pallet::<T>::generate_task_id(account_id.clone(), vec![0]);
+			let task = OldTask::<T> {
 				owner_id: account_id.clone(),
 				provided_id: vec![1],
 				execution_times: vec![10].try_into().unwrap(),
@@ -519,7 +519,7 @@ benchmarks! {
 					frequency: 11,
 				},
 			};
-			OldAccountTasks::<T>::insert(account_id.clone(), task_id_1, task_1);
+			OldAccountTasks::<T>::insert(account_id.clone(), task_id, task);
 		}
 	}: { AddScheduleToTask::<T>::on_runtime_upgrade() }
 	verify {
--- a/pallets/automation-time/src/benchmarking.rs
+++ b/pallets/automation-time/src/benchmarking.rs
@@ -497,4 +497,32 @@ benchmarks! {
 
 		schedule_notify_tasks::<T>(caller.clone(), vec![new_time_slot], T::MaxTasksPerSlot::get());
 	}: { AutomationTime::<T>::shift_missed_tasks(last_time_slot, diff) }
+
+	migration_v5 {
+		let v in 1..100;
+
+		use migrations::*;
+		use frame_support::traits::OnRuntimeUpgrade;
+
+		for i in 0..v {
+			let account_id: T::AccountId = account("Account", 0, i);
+			let task_id_1 = Pallet::<T>::generate_task_id(account_id.clone(), vec![0]);
+			let task_1 = v5::OldTask::<T> {
+				owner_id: account_id.clone(),
+				provided_id: vec![1],
+				execution_times: vec![10].try_into().unwrap(),
+				executions_left: 1,
+				action: v5::OldAction::<T>::AutoCompoundDelegatedStake {
+					delegator: account_id.clone(),
+					collator: account_id.clone(),
+					account_minimum: 100u32.into(),
+					frequency: 11,
+				},
+			};
+			v5::AccountTasks::<T>::insert(account_id.clone(), task_id_1, task_1);
+		}
+	}: { v5::MigrateToV5::<T>::on_runtime_upgrade() }
+	verify {
+		assert_eq!(v, crate::AccountTasks::<T>::iter().count() as u32);
+	}
 }
--- a/pallets/automation-time/src/benchmarking.rs
+++ b/pallets/automation-time/src/benchmarking.rs
@@ -49,14 +49,14 @@ fn schedule_notify_tasks<T: Config>(owner: T::AccountId, times: Vec<u64>, count:
 
 	for _ in 0..count {
 		provided_id = increment_provided_id(provided_id);
-		let task = TaskOf::<T>::create_event_task(
+		let task = TaskOf::<T>::create_event_task::<T>(
 			owner.clone(),
 			provided_id.clone(),
-			times.clone().try_into().unwrap(),
+			times.clone(),
 			vec![4, 5, 6],
-		);
-		let task_id =
-			AutomationTime::<T>::schedule_task(&task, provided_id.clone(), times.clone()).unwrap();
+		)
+		.unwrap();
+		let task_id = AutomationTime::<T>::schedule_task(&task, provided_id.clone()).unwrap();
 		<AccountTasks<T>>::insert(owner.clone(), task_id, task);
 	}
 
@@ -72,15 +72,15 @@ fn schedule_transfer_tasks<T: Config>(owner: T::AccountId, time: u64, count: u32
 
 	for _ in 0..count {
 		provided_id = increment_provided_id(provided_id);
-		let task = TaskOf::<T>::create_native_transfer_task(
+		let task = TaskOf::<T>::create_native_transfer_task::<T>(
 			owner.clone(),
 			provided_id.clone(),
-			vec![time].try_into().unwrap(),
+			vec![time],
 			recipient.clone(),
 			amount.clone(),
-		);
-		let task_id =
-			AutomationTime::<T>::schedule_task(&task, provided_id.clone(), vec![time]).unwrap();
+		)
+		.unwrap();
+		let task_id = AutomationTime::<T>::schedule_task(&task, provided_id.clone()).unwrap();
 		<AccountTasks<T>>::insert(owner.clone(), task_id, task);
 	}
 
@@ -100,17 +100,17 @@ fn schedule_xcmp_tasks<T: Config>(owner: T::AccountId, times: Vec<u64>, count: u
 
 	for _ in 0..count {
 		provided_id = increment_provided_id(provided_id);
-		let task = TaskOf::<T>::create_xcmp_task(
+		let task = TaskOf::<T>::create_xcmp_task::<T>(
 			owner.clone(),
 			provided_id.clone(),
-			times.clone().try_into().unwrap(),
+			times.clone(),
 			para_id.clone().try_into().unwrap(),
 			T::GetNativeCurrencyId::get(),
 			vec![4, 5, 6],
 			5_000,
-		);
-		let task_id =
-			AutomationTime::<T>::schedule_task(&task, provided_id.clone(), times.clone()).unwrap();
+		)
+		.unwrap();
+		let task_id = AutomationTime::<T>::schedule_task(&task, provided_id.clone()).unwrap();
 		<AccountTasks<T>>::insert(owner.clone(), task_id, task);
 	}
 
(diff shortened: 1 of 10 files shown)


pallets/automation-time/src/migrations.rs
```diff
@@ -438,6 +437,163 @@ pub mod v4 {
...
+				target: "automation-time",
+				"migration: AutomationTime storage version v5 migration succesful! Migrated {} tasks.",
+				migrated_tasks
+			);
+
+			// 1 read + write to transform each task
+			let weight = T::DbWeight::get().reads_writes(migrated_tasks, migrated_tasks);
+
+			// Adding a buffer for the rest of the code
+			weight + (migrated_tasks * 10_000_000)
```
nvengal, line 575: Need to write a benchmark to prove the extra weight is correct

pallets/automation-time/src/lib.rs
```diff
@@ -1228,10 +1221,10 @@ pub mod pallet {
 		pub fn schedule_task(
 			task: &TaskOf<T>,
 			provided_id: Vec<u8>,
-			execution_times: Vec<UnixTime>,
 		) -> Result<TaskId<T>, Error<T>> {
 			let owner_id = task.owner_id.clone();
 			let task_id = Self::generate_task_id(owner_id.clone(), provided_id.clone());
+			let execution_times = task.execution_times(vec![]);
```
rhuttman, line 1227: I only ever see you pass in `vec![]` here. Why does `task.execution_times` need `Schedule::Recurring` to append to a vector anyway?

pallets/automation-time/src/lib.rs
```diff
@@ -1279,24 +1272,17 @@ pub mod pallet {
...
-			for time in execution_times.iter() {
-				Self::is_valid_time(*time)?;
-			}
+			schedule.valid::<T>()?;
+			let executions = schedule.number_of_known_executions();
 
 			// Execution fee
-			let exeuction_fee =
-				Self::calculate_execution_fee(&action, execution_times.len().try_into().unwrap())?;
+			let exeuction_fee = Self::calculate_execution_fee(&action, executions)?;
```
rhuttman, line 1285: `exeuction_fee ` -> `execution`. Always fun finding my old typos

pallets/automation-time/src/lib.rs
```diff
@@ -1343,27 +1325,74 @@ pub mod pallet {
...
+		fn reschedule_or_remove_task(
 			task_id: TaskId<T>,
-			task: &TaskOf<T>,
-			execution_times: Vec<UnixTime>,
-		) -> Result<(), DispatchError> {
-			let new_executions = execution_times.len().try_into().unwrap();
+			task: &mut TaskOf<T>,
+			dispatch_error: Option<DispatchError>,
+		) {
+			let remove_on_err = |task: &TaskOf<T>, err: DispatchError| {
```
rhuttman, line 1333: this seems like a world where we want to take ownership of `Task`. Also why an in-line fcn and not a real fcn?

pallets/automation-time/src/lib.rs
```diff
@@ -1343,27 +1325,74 @@ pub mod pallet {
...
+				Self::deposit_event(Event::<T>::TaskNotRescheduled {
+					who: task.owner_id.clone(),
+					task_id,
+					error: err,
```
rhuttman, line 1337: Would it make more sense for this to be a `scheduling_error`? The error from the action has already been sent in an event, so this would either be a dupe error or a scheduling error. The fact that it could be either or without an easy way to distinguish could also be confusing.

pallets/automation-time/src/lib.rs
```diff
@@ -1343,27 +1325,74 @@ pub mod pallet {
...
+					Err(e) => {
+						remove_on_err(task, e);
+					},
+				};
+			}
+		}
 
-			Self::insert_scheduled_tasks(task_id, task, execution_times.clone())?;
+		fn reschedule_existing_task(task_id: TaskId<T>, task: &mut TaskOf<T>) -> DispatchResult {
+			match task.schedule {
```
rhuttman, line 1358: I think you can use an `if let` here

pallets/automation-time/src/lib.rs
```diff
@@ -1343,27 +1325,74 @@ pub mod pallet {
...
+				Schedule::Recurring { ref mut next_execution_time, frequency } => {
+					let new_execution_time = next_execution_time
```
rhuttman, line 1360: shouldn't this logic also live on the struct?

pallets/automation-time/src/lib.rs
```diff
@@ -1343,27 +1325,74 @@ pub mod pallet {
...
+						.checked_add(frequency)
+						.ok_or(Error::<T>::InvalidTime)?;
+					*next_execution_time = new_execution_time;
 
-			T::FeeHandler::withdraw_fee(&task.owner_id, fee.clone())
-				.map_err(|_| Error::<T>::LiquidityRestrictions)?;
+					// TODO: should execution fee depend on whether task is recurring?
+					let fee = Self::calculate_execution_fee(&task.action, 1)?;
```
rhuttman, line 1366: this doesn't take the XCMP fees into account

rhuttman (changes requested):

pallets/automation-time/src/lib.rs
```diff
@@ -304,6 +304,15 @@ pub mod pallet {
 			who: AccountOf<T>,
 			task_id: TaskId<T>,
 		},
+		TaskRescheduled {
```
lreesby, line 307: can you add a description?  `rescheduled` is kind of a confusing way to put it.  not sure I have a much better idea.  maybe `renewed`?   unless I'm misunderstanding what the event is for.

pallets/automation-time/src/types.rs
```diff
@@ -83,20 +82,83 @@ impl<AccountId: Clone + Decode, Balance: AtLeast32BitUnsigned, CurrencyId: Defau
...
+				Schedule::Recurring { next_execution_time: t2, frequency: f2 },
+			) => t1 == t2 && f1 == f2,
+			_ => false,
+		}
+	}
+}
+
+impl<MaxExecutionTimes: Get<u32>> Schedule<MaxExecutionTimes> {
+	pub fn new_fixed_schedule<T: Config>(execution_times: Vec<UnixTime>) -> Result<Self, Error<T>> {
+		let mut execution_times = execution_times.clone();
```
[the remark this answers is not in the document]
nvengal: You're very right. Good catch.

pallets/automation-time/src/types.rs
```diff
@@ -83,20 +82,83 @@ impl<AccountId: Clone + Decode, Balance: AtLeast32BitUnsigned, CurrencyId: Defau
...
+		let execution_times: BoundedVec<UnixTime, MaxExecutionTimes> =
+			execution_times.try_into().map_err(|_| Error::<T>::TooManyExecutionsTimes)?;
+		Ok(Self::Fixed { execution_times, executions_left })
+	}
+
+	pub fn new_recurring_schedule(next_execution_time: UnixTime, frequency: Seconds) -> Self {
+		Self::Recurring { next_execution_time, frequency }
+	}
+
+	pub fn valid<T: Config>(&self) -> Result<(), DispatchError> {
```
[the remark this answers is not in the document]
nvengal: This is a reasonable suggestion. I changes the pattern around testing a little as now tasks that are created with the helper methods require valid times (ex. no more `executions_time=vec![0]`). That change seems reasonable to me so I went ahead and made this update.

pallets/automation-time/src/types.rs
```diff
@@ -125,63 +184,80 @@ impl<AccountId: Clone, Balance, CurrencyId, MaxExecutionTimes: Get<u32>>
...
 			account_minimum,
-			frequency,
 		};
-		Self::new(owner_id, provided_id, vec![execution_time].try_into().unwrap(), action)
+		let schedule =
+			Schedule::<MaxExecutionTimes>::new_recurring_schedule(next_execution_time, frequency);
+		Ok(Self::new(owner_id, provided_id, schedule, action))
+	}
+
+	pub fn execution_times(&self, mut l: Vec<UnixTime>) -> Vec<UnixTime> {
```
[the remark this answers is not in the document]
nvengal: Good catch! this was a leftover param from when I was struggling with rust's lifetime rules

pallets/automation-time/src/types.rs
```diff
@@ -125,63 +184,80 @@ impl<AccountId: Clone, Balance, CurrencyId, MaxExecutionTimes: Get<u32>>
...
+			Schedule::Fixed { execution_times, .. } => execution_times.to_vec(),
+			Schedule::Recurring { next_execution_time, .. } => {
+				l.push(*next_execution_time);
+				l
+			},
+		}
+	}
+
+	pub fn executions_left(&self) -> u32 {
+		*&self.schedule.number_of_known_executions()
```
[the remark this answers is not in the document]
nvengal: Confused by this. What do you mean?

pallets/automation-time/src/lib.rs
```diff
@@ -357,7 +366,7 @@ pub mod pallet {
 				Action::Notify { message },
 				who,
 				provided_id,
-				execution_times,
+				Schedule::<T::MaxExecutionTimes>::new_fixed_schedule::<T>(execution_times)?,
```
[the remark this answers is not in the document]
nvengal: No reason. Updated it to match the rest.

pallets/automation-time/src/lib.rs
nvengal: Leftover code from my struggle with rust's lifetime rules. Removed now. Thanks for flagging.

pallets/automation-time/src/lib.rs
nvengal: Re: inline function. I don't have a good reason here... Moved it to a regular function.

Re: taking ownership. This seems reasonable to me. I don't know if there's a better way to have done this so linking commit here for easier review: [147fa23](https://github.com/OAK-Foundation/OAK-blockchain/pull/282/commits/147fa2398af57a9dd51dc09d993e69bf78d30f62)

I don't know if/how to take ownership of the reference so I changed the upstream function to pass the owned value down.

pallets/automation-time/src/lib.rs
nvengal: That's a reasonable concern. I've opted to add a different event when we don't reschedule for task failure (TaskNotRescheduled) vs try to reschedule but fail (TaskFailedToReschedule). I could really go either way on distinct event vs error. Let me know what you think.

pallets/automation-time/src/lib.rs
nvengal: I opted for a `match` instead of an `if let` because I wanted to demonstrate that this code was exhaustive on all permutations of `Schedule` in the future if we add a new variant to `Schedule` the compiler will flag that the new variant should specify if/how it should be rescheduled. We'll lose that nice compiler feature if this is converted to an `if let`.

pallets/automation-time/src/lib.rs
nvengal: I think that's reasonable. I didn't originally include it because it requires passing the `task` into the method and dependency felt a little too circular. I think I'd want to split this across the Task and Schedule struct to make that a bit cleaner. Will think on this.

pallets/automation-time/src/lib.rs
nvengal: I don't have a good grasp on xcmp fees yet. How are you thinking that I should include the xcmp fees here?

pallets/automation-time/src/lib.rs
nvengal: I added a doc comment to help clear it up. Let me know what you think. I'm open to renaming it, but  also not sure on a good name. I'm not sure on `renew` will think on it.

pallets/automation-time/src/lib.rs
```diff
@@ -519,8 +525,10 @@ pub mod pallet {
 
 			let encoded_call = call.encode();
 			let action = Action::DynamicDispatch { encoded_call };
+			let schedule =
```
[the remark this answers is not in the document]
nvengal: Yup, that's what's coming in the next pr!

pallets/automation-time/src/migrations.rs
```diff
@@ -438,6 +437,159 @@ pub mod v4 {
 	}
 }
```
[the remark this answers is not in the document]
nvengal: yup doing that here, https://github.com/OAK-Foundation/OAK-blockchain/pull/295

pallets/automation-time/src/lib.rs
```diff
@@ -1343,27 +1325,74 @@ pub mod pallet {
...
 
+		fn handle_task_post_processing(
+			task_id: TaskId<T>,
+			task: &mut TaskOf<T>,
+			error: Option<DispatchError>,
+		) {
+			match task.schedule {
+				Schedule::Fixed { .. } =>
+					Self::decrement_task_and_remove_if_complete(task_id, task),
+				Schedule::Recurring { .. } => Self::reschedule_or_remove_task(task_id, task, error),
```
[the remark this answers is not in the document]
nvengal: This is a reasonable suggestion. Ryan suggested something along the same lines. I'll create a pr of what that might look like and we can decide if that's the path we want to move down.

pallets/automation-time/src/lib.rs
nvengal: Splitting this suggestion into a new pr so we can discuss the pros/cons in isolation.

pallets/automation-time/src/types.rs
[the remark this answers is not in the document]
rhuttman: I meant why not just ask the schedule directly. Why use `task. executions_left` vs `task.schedule.executions_left`?

pallets/automation-time/src/types.rs
```diff
@@ -323,4 +414,199 @@ mod tests {
...
+						SCHEDULED_TIME + 3600,
+						SCHEDULED_TIME + 3650
+					]),
+					Error::<Test>::InvalidTime
+				);
+			})
+		}
+
+		#[test]
+		fn checks_for_recurring_schedule_validity() {
```
rhuttman, line 555: I don't think you covered the case where the `frequency` is not a valid number. You where the frequency is 0, but not where it's say 3650.

pallets/automation-time/src/lib.rs
rhuttman: we just need to collect it here. The xcmp fees are separate from the execution. Right now this will cause someone's xcmp action to fail because there won't be enough funds in the account to pay for it. You can see how this is done in `validate_and_schedule_task` which you already updated. Might be worth trying to refactor to use the same fee logic.

Additionally I am now agreeing that we should factor this weight into the execution fee, but that can come later.

pallets/automation-time/src/benchmarking.rs
```diff
@@ -496,4 +497,32 @@ benchmarks! {
...
+
+	migration_add_schedule_to_task {
+		let v in 1..100;
+
+		use migrations::add_schedule_to_task::{AccountTasks as OldAccountTasks, AddScheduleToTask, OldAction, OldTask};
+		use frame_support::traits::OnRuntimeUpgrade;
+
+		for i in 0..v {
+			let account_id: T::AccountId = account("Account", 0, i);
+			let task_id_1 = Pallet::<T>::generate_task_id(account_id.clone(), vec![0]);
```
rhuttman, line 509: did you mean to have a `task_id_2`?

rhuttman (changes requested):

pallets/automation-time/src/benchmarking.rs
nvengal: I did originally, but ended up removing it. Putting in two tasks per `v` completely screws up the benchmarking numbers.

justinzhou93 (approved):
my comments were addressed, so looks good to me. seems like still a few pending comments from Ryan, so I'd make sure he's on the same page, but otherwise lgtm on my concerns.

pallets/automation-time/src/types.rs
nvengal: Good catch. Added a assertion for this.

pallets/automation-time/src/types.rs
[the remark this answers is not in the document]
nvengal: Removed the unnecessary method.

pallets/automation-time/src/benchmarking.rs
nvengal: Updated the var names to remove confusion

pallets/automation-time/src/lib.rs
nvengal: Rather than make this pr larger, I broke the structural changes to fee payment into its own pr: https://github.com/OAK-Foundation/OAK-blockchain/pull/298
Will wait on merging the fee pr before moving forward with this.

justinzhou93 (approved):
re lgtm - should have addressed rest of Ryan and Laura's concerns from what i could see

chrisli30 (approved):
LGTM

PR #283, opened 2022-09-15
Adding OAK Initial Allocation & Vesting 

irsal:
**OAK's Tokenomics can be found here: https://oak.tech/oak/token/**

## Notes
- oak_alloc.json is for the intial allocation upon launch/block production.  Anything after that is in oak_vesting.json
- Timestamps are all unix timestamps. Set for the the 1st of the month at 12PM (GMT-8). aka 12PM PST / California time.
- The unixtimestamp may be changed at a later time due to TDE terms.
- We need to mint (aka vest) the day before distribution. So if we distribute to different "individual wallets" at the 6th of the month, we need to mint them before that date in the "company wallets".
- Folks who have wallets in the sheets need to double check and verify the address. 
- JSON values are in PLANCK units, which is the smallest possible OAK unit. This means 10 decimal spots for 1 unit, or 1E10. `20000000000` in the json is equivalent to 2 OAK.
- Given the bigint conversion, we extended all values by adding 10 zeros. (For example, if the value is 4166666 OAK, we converted this to 41666660000000000 planck)

## Totals

(Computed via Rust.)
- oak_vesting.json ==> `7599999950000000000` planck 
- oak_alloc.json      ==> `80230000000000000` planck
- Sum of both             ==> `7680229950000000000` planck

## Engineering Testing

- [x] Total amounts make sense
- [x] No duplicate timestamps 
- [x]  No duplicate addresses per time stamp
- [x] Same diff for consecutive timestamps (~1 month) --> JS
- [x] Output of totals per month 
- [x] Output of totals per address 
- [x] Output of totals per address at 6th, 12th, 18th, 24, 30th, and 36th month
- [x] [output.txt](https://github.com/OAK-Foundation/OAK-blockchain/files/9579640/output.txt)

Screen Shot 2022-09-15 at 1 11 11 PM


distribution/oak_vesting.json
```diff
@@ -106,7 +106,7 @@
     ]
   ],
   [
-    1701457200,
+    1701460800,
```
justinzhou93, line 109: dates changed because they're supposed to be on the 1st of each month at noon PST. However, some of them were different, at 11 AM PST instead. This was probably due to daylight savings because the timestamps could have been based on UTC time. These have now all been adjusted to 12 PM PST on the 1st of every month for 4 years

chrisli30: Yes, I converted them back with moment() and verified that.

chrisli30 (approved):
I converted this json file back to csv and verified that it’s the same as 
https://docs.google.com/spreadsheets/d/1-LWgg9Q27VCYpnIOc5gAUHPGTNFxJa9h_TXguVVGYZ8/edit#gid=646269391

Great job!

PR #287, opened 2022-09-15
Re-add schedule xcmp benchmarking range

nvengal:
Accidentally removed this range because I don't know how benchmarks work!


lreesby (approved):
lgtm

PR #292, opened 2022-09-16
Add Zombies

lreesby:


zombienets/neumann/single-chain.toml
```diff
@@ -0,0 +1,24 @@
+[settings]
+provider = "native"
+timeout = 1000
+
+[relaychain]
+default_command = "../polkadot/target/release/polkadot"
+chain = "rococo-local"
```
rhuttman, line 7: this should probably be `westend-local` as we need to deploy neumann to purestake alphanet which uses westend

rhuttman (approved):

PR #295, opened 2022-09-19
Cleanup old migrations

nvengal:
Experiment with migrations::old mods to store already-run migrations.
Migrations::old is purposefully not added to module tree
to avoid unnecessary complier errors and warnings.


rhuttman (approved):
lgtm on the code. Let's do a quick check-in with the team before we merge

PR #298, opened 2022-09-26
Add a stateful FeeHandler to de-dup execution fee payment code

nvengal:
Based on this comment: https://github.com/OAK-Foundation/OAK-blockchain/pull/282#discussion_r978926708


pallets/automation-time/src/fees.rs
```diff
@@ -16,36 +16,57 @@
 // limitations under the License.
 
 /// ! Traits and default implementation for paying execution fees.
```
nvengal, line 18: Attempting to choose between two implementations:
1. FeeHandler allows explicitly building and executing fee payment
https://github.com/OAK-Foundation/OAK-blockchain/compare/5830b71c7aacf48..0ddc6fd9bcd7d
2. FeeHandler wraps a closure and implicitly checks and charges fees
https://github.com/OAK-Foundation/OAK-blockchain/compare/0ddc6fd9bcd7d..ce27075e67881

Looking for feedback on folks preferred implementation before I spend time adding tests. I lean towards the second implementation since it has the least duplicated code and clearest separation of concerns between the Pallet code and the FeeHandler (in the second implementation the developer in the Pallet struct doesn't need to remember to call `pay_fee` at the appropriate time). However, passing the closure is a more complex implementation so maybe it's not worth cost.

justinzhou93: i tend to agree with the closure implementation as well. It keeps the code cleaner, such that the "prework" code that should live in the lib file still lives there, but all of the logic around fee calculation and execution is pulled into the fees.rs file. and for the "prework" code that needs to execute in between the calculation and charging of the fee is non-static and can be defined by the function that's calling in. funny enough, it's actually kinda like React, where you can inject a function from parent to child components and the function keeps its original context (the parent component), unless you bind it to the child component to now give it new context. this way, you can define functionality in the controller component and keep the children component "dumber" (i.e. simpler) and stateless. fun to see that interaction cross languages. very cool!

justinzhou93 (approved):

pallets/automation-time/src/mock.rs
```diff
@@ -323,3 +324,121 @@ pub fn new_test_ext(state_block_time: u64) -> sp_io::TestExternalities {
 	ext.execute_with(|| Timestamp::set_timestamp(state_block_time));
 	ext
 }
+
+pub fn schedule_task(
```
nvengal, line 328: Move these methods from the `tests` module to `mocks` to make it easier to reuse across different test submodules.

pallets/automation-time/src/fees.rs
```diff
@@ -148,3 +148,78 @@ where
...
+			let alice = AccountId32::new(ALICE);
+			get_funds(alice.clone());
+			let starting_funds = Balances::free_balance(alice.clone());
+			let _ = <Test as crate::Config>::FeeHandler::pay_checked_fees_for(
+				&alice,
+				&Action::Notify { message: vec![0] },
+				1,
+				|| Ok(()),
+			);
+			assert!(starting_funds > Balances::free_balance(alice))
```
justinzhou93, line 205: could this test be combined with executes_prereq_function? like basically having 1 success test case and we check the state of fee and prereq trigger after that

justinzhou93 (commented):
1 nit/question otherwise lgtm

justinzhou93 (approved):

PR #299, opened 2022-09-29
Fix compiler warnings

nvengal:
Compiler warnings for unused imports make it difficult to recognize when a new issue is introduced.

### DIFF
--- a/pallets/xcmp-handler/src/lib.rs
+++ b/pallets/xcmp-handler/src/lib.rs
@@ -387,12 +387,10 @@ pub mod pallet {
 			para_id: ParachainId,
 			target_instructions: xcm::v2::Xcm<()>,
 		) -> Result<(), DispatchError> {
-			let destination_location: Junctions =
-				if cfg!(feature = "runtime-benchmarks") && cfg!(not(test)) {
-					Here
-				} else {
-					X1(Junction::Parachain(para_id.into()))
-				};
+			let destination_location = Junction::Parachain(para_id.into());
+
+			#[cfg(all(not(test), feature = "runtime-benchmarks"))]
+			let destination_location: Junctions = Here;
 
 			// Send to target chain
 			T::XcmSender::send_xcm((1, destination_location), target_instructions).map_err(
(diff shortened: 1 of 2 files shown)


justinzhou93 (approved):

pallets/xcmp-handler/src/lib.rs
```diff
@@ -387,10 +387,12 @@ pub mod pallet {
...
-			let destination_location = Junction::Parachain(para_id.into());
-
-			#[cfg(all(not(test), feature = "runtime-benchmarks"))]
-			let destination_location: Junctions = Here;
+			let destination_location: Junctions =
+				if cfg!(feature = "runtime-benchmarks") && cfg!(not(test)) {
+					Here
+				} else {
+					X1(Junction::Parachain(para_id.into()))
+				};
```
nvengal, line 395: @justinzhou93 I'm not sure about this change. I think that the my code is equivalent, but honestly I don't have enough confidence in the tests here or my knowledge of xcm to be sure about that. Flagging to see if you have a better understanding of this. If not, I'll revert this change as having low value vs risk.

justinzhou93: also not 100% confidence on my side here, but i do believe the if/else should provide the same functionality. although the destination location seems to be a tax different in your re-write. 
`Junction::Parachain(para_id.into());` seems different from `X1(Junction::Parachain(para_id.into()))`. extra X1 wrapper at least. 

although at least personally i do actually prefer the compile flag cause i think we do it in quite a few other places, whereas the if/else feels more like non-Rust implementation. personally i actually think it'd be better split out as a separate function that will provide destination location (regardless of how it's written) so that at least the reader doesn't have to consider all that messiness in this function and can have separate function for determining destination

nvengal: Going to go ahead and revert this change.
The underlying issue is that `Here` is a different type than `Junction::Parachain`. Wrapping `Junction::Parachain` in the `X1` is what allows getting past this compiler warning, but it sounds like neither one of us is ready to make that change yet.

justinzhou93 (approved):

PR #300, opened 2022-10-05
Add scheduling options to dynamic dispatch and xcmp tasks

nvengal:
Can schedule as either `Fixed` (runs for each specified execution time)
image
or `Recurring` (runs until an error occurs or the wallet runs out of funds)
image

- [ ] Depends on https://github.com/OAK-Foundation/OAK-blockchain/pull/301

PR #301, opened 2022-10-05
Remove generic param from Schedule enum

nvengal:
Simplifies the Schedule enum removing the generic `MaxExecutionTimes` arg used for storing execution times as a `BoundedVec`. Instead, in this pr, execution_times are stored as a `Vec` and number of executions is checked for validity in the `valid` method.

### DIFF
--- a/pallets/automation-time/src/types.rs
+++ b/pallets/automation-time/src/types.rs
@@ -115,7 +115,7 @@ impl Schedule {
 		mut execution_times: Vec<UnixTime>,
 	) -> Result<Self, DispatchError> {
 		Pallet::<T>::clean_execution_times_vector(&mut execution_times);
-		let executions_left = execution_times.len();
+		let executions_left = execution_times.len() as u32;
 		let schedule = Self::Fixed { execution_times, executions_left };
 		schedule.valid::<T>()?;
 		Ok(schedule)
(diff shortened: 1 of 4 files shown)


pallets/automation-time/src/types.rs
```diff
@@ -87,37 +87,18 @@ impl<AccountId: Clone + Decode, Balance: AtLeast32BitUnsigned, CurrencyId: Defau
 	}
 }
 
-#[derive(Debug, Encode, Decode, TypeInfo)]
-#[scale_info(skip_type_params(MaxExecutionTimes))]
-pub enum Schedule<MaxExecutionTimes: Get<u32>> {
```
nvengal, line 92: There is an argument to be made that encoding the MaxExecutionTimes in the type is a better implementation since its a stronger type bound and maybe less prone to error. However, in order to use the `Schedule` enum at the api boundary when scheduling tasks, the enum needs to be entirely serializable. `T::MaxExecutionTimes` does not currently fit that requirement and I don't think the work required to make it fit is worth the cost.

PR in question that adds `Schedule` to the api boundary- https://github.com/OAK-Foundation/OAK-blockchain/pull/300

pallets/automation-time/src/types.rs
```diff
@@ -87,37 +87,18 @@ impl<AccountId: Clone + Decode, Balance: AtLeast32BitUnsigned, CurrencyId: Defau
...
-	Fixed { execution_times: BoundedVec<UnixTime, MaxExecutionTimes>, executions_left: u32 },
+#[derive(Debug, Encode, Decode, PartialEq, TypeInfo)]
+pub enum Schedule {
+	Fixed { execution_times: Vec<UnixTime>, executions_left: u32 },
 	Recurring { next_execution_time: UnixTime, frequency: Seconds },
 }
 
-impl<B: Get<u32>> PartialEq for Schedule<B> {
```
nvengal, line 97: Getting rid of the generic parameter allows us derive `PartialEq`.

pallets/automation-time/src/types.rs
chrisli30: Yes I agree. The delta is not big enough for the serialization of enum work.

chrisli30 (approved):
Looks straight-forward to me 👍

PR #302, opened 2022-10-07
Schedule is invalid with no execution times

nvengal:
This bug allowed users to submit tasks with no execution_times that would sit in storage forever.
Tasks are only cleaned from storage once they are executed.


chrisli30 (approved):
This PR is still in Draft but it looks like having solved the no-exe-time problem?

PR #303, opened 2022-10-14
Benchmark dynamic dispatch

nvengal:


pallets/automation-time/src/benchmarking.rs
```diff
@@ -526,4 +585,10 @@ benchmarks! {
 	verify {
 		assert_eq!(v, crate::AccountTasks::<T>::iter().count() as u32);
 	}
+
+	impl_benchmark_test_suite!(
```
nvengal, line 589: Because this macro was missing the benchmarks were not being run as part of the unit tests. Since they weren't running in the unit tests some of these benchmarks have been broken since the last release and we haven't noticed. Fixes for the broken benchmarks are included in this pr.

chrisli30: noice!

chrisli30 (approved):
Yep, looks good to me!

PR #314, opened 2022-12-02
[Migration] Add Mangata TUR chain currency combo for xcm fees

nvengal:


pallets/xcmp-handler/src/migrations/add_mgx_tur_chain_currency_combo.rs
```diff
@@ -0,0 +1,50 @@
...
+	traits::{Get, OnRuntimeUpgrade},
+	weights::Weight,
+};
+
+const MANGATA_PARA_ID: u32 = 2110;
+const FEE_PER_SECOND: u128 = 5_376_000_000_000;
+
+pub struct AddMgxTurChainCurrencyCombo<T>(PhantomData<T>);
+impl<T: Config> OnRuntimeUpgrade for AddMgxTurChainCurrencyCombo<T> {
+	fn on_runtime_upgrade() -> Weight {
```
nvengal, line 14: Successfully testing this migration using `try-runtime`
```
2022-12-02 15:42:06.928  INFO                 main xcmp-handler: migration: AddMgxTurChainCurrencyCombo POST migration checks succesful!
```

chrisli30 (approved):
LGTM! One thing occurs to me is that we need to mark this PR as "Migration", so when we batch all for the next upgrade we do not forget about the migration.

We can add [Migration] to the title of the PR and add a tag to it.

PR #319, opened 2022-12-13
Add Shibuya asset

imstar15:
image

### DIFF
--- a/node/src/chain_spec/turing.rs
+++ b/node/src/chain_spec/turing.rs
@@ -121,7 +121,7 @@ pub fn turing_development_config() -> ChainSpec {
 								decimals: 18,
 								name: b"Rocstar".to_vec(),
 								symbol: b"RSTR".to_vec(),
-								existential_deposit: Default::default(),
+								existential_deposit: 10_000_000_000_000_000,
 								location: Some(MultiLocation::new(1, X1(Parachain(2006))).into()),
 								additional: CustomMetadata {
 									fee_per_second: Some(416_000_000_000),
@@ -137,7 +137,7 @@ pub fn turing_development_config() -> ChainSpec {
 								decimals: 18,
 								name: b"Shiden".to_vec(),
 								symbol: b"SDN".to_vec(),
-								existential_deposit: Default::default(),
+								existential_deposit: 10_000_000_000_000_000,
 								location: Some(MultiLocation::new(1, X1(Parachain(2007))).into()),
 								additional: CustomMetadata {
 									fee_per_second: Some(416_000_000_000),
--- a/node/src/chain_spec/turing.rs
+++ b/node/src/chain_spec/turing.rs
@@ -11,8 +11,8 @@ use codec::Encode;
 use common_runtime::constants::currency::{DOLLAR, TOKEN_DECIMALS};
 use primitives::{assets::CustomMetadata, AccountId, AuraId, Balance, TokenId};
 use turing_runtime::{
-	AssetRegistryConfig, CouncilConfig, PolkadotXcmConfig, TechnicalMembershipConfig, ValveConfig,
-	VestingConfig, XcmpHandlerConfig, Parachain, X1,
+	AssetRegistryConfig, CouncilConfig, Parachain, PolkadotXcmConfig, TechnicalMembershipConfig,
+	ValveConfig, VestingConfig, XcmpHandlerConfig, X1,
 };
 use xcm::{
 	opaque::latest::{Junctions::Here, MultiLocation},
@@ -207,14 +207,12 @@ fn testnet_genesis(
 							decimals: 18,
 							name: b"Mangata".to_vec(),
 							symbol: b"MGX".to_vec(),
+							existential_deposit: Default::default(),
+							location: Some(MultiLocation::new(1, X1(Parachain(2110))).into()),
 							additional: CustomMetadata {
 								fee_per_second: Some(416_000_000_000),
 								conversion_rate: None,
 							},
-							existential_deposit: Default::default(),
-							location: Some(
-								MultiLocation::new(1, X1(Parachain(2110))).into(),
-							),
 						},
 					),
 				),
@@ -225,14 +223,12 @@ fn testnet_genesis(
 							decimals: 18,
 							name: b"Rocstar".to_vec(),
 							symbol: b"RSTR".to_vec(),
+							existential_deposit: Default::default(),
+							location: Some(MultiLocation::new(1, X1(Parachain(2006))).into()),
 							additional: CustomMetadata {
 								fee_per_second: Some(416_000_000_000),
 								conversion_rate: None,
 							},
-							existential_deposit: Default::default(),
-							location: Some(
-								MultiLocation::new(1, X1(Parachain(2006))).into(),
-							),
 						},
 					),
 				),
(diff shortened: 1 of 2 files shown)


node/src/chain_spec/turing.rs
```diff
@@ -183,23 +183,43 @@ fn testnet_genesis(
...
-					},
+					),
+				),
+				(
+					1,
+					orml_asset_registry::AssetMetadata::<Balance, CustomMetadata>::encode(
+						&orml_asset_registry::AssetMetadata {
+							decimals: 18,
+							name: b"Shibuya".to_vec(),
+							symbol: b"SBY".to_vec(),
```
chrisli30, line 209: hmm what token symbol to add here is a little tricky. 

First, we are not testing aginst Shibuya, but Rocstar.
Rococo - Rocstar - RSTR
Kusama - Shiden - SDN

Second, is it okay that we use the symbol SDN here, and map it to RSTR for Staging and actual SDN on Turing?

I’ll defer that to Nikhil.

node/src/chain_spec/turing.rs
```diff
@@ -183,23 +183,43 @@ fn testnet_genesis(
...
+								fee_per_second: Some(416_000_000_000),
+								conversion_rate: None,
+							},
+							existential_deposit: Default::default(),
+							location: Some(
+								MultiLocation::new(1, X1(Parachain(2000))).into(),
+							),
+						},
+					),
 				),
```
chrisli30, line 220: Since we are doing this, maybe add MGX as well?

chrisli30 (commented):
Added two comments for @nvengal to take a look.

node/src/chain_spec/turing.rs
nvengal: Since we're adding tokens that are used exclusively for `development` here, `additonal_assets` should be added as a parameter for to this function `testnet_genesis` and be passed down from the caller. That way we can be explicit about tokens for dev, turing-staging, turing. The `Native` token can stay embedded here since all chains will have the same one.

> Second, is it okay that we use the symbol SDN here, and map it to RSTR for Staging and actual SDN on Turing?

If the same token was used for `turing-staging` and `turing` then we could use a migration here. Since they both have different token symbols, it will need to be a governance proposal. We will need to be careful to make sure that both tokens end up with the same currency id since the integer value of the currency id is referenced in the code.

node/src/chain_spec/turing.rs
nvengal: Yup, that is reasonable.

runtime/turing/src/lib.rs
```diff
@@ -36,7 +36,7 @@ use sp_runtime::{
 	transaction_validity::{TransactionSource, TransactionValidity},
 	AccountId32, ApplyExtrinsicResult, DispatchError, ModuleError, Percent, RuntimeDebug,
 };
-use xcm::{
+pub use xcm::{
```
nvengal, line 39: Why is this `pub`?

node/src/chain_spec/turing.rs
nvengal: Also, we should check if `Rocstar` and `Shiden` have the same parachain id. If not, it may make more sense to treat them distinctly.

nvengal (changes requested):

node/src/chain_spec/turing.rs
chrisli30: Good call. No, they have different paraId, so we will need to treat them differently.
Rocstar: 2006
Shiden: 2007

runtime/turing/src/lib.rs
imstar15: Since xcm::v1::Junction::Parachain and xcm::latest::prelude::X1 will be used in node/src/chain_spec/turing.rs, they need to be exposed.

This is similar to mangata's code.

https://github.com/mangata-finance/mangata-node/blob/25b2af6d95de3399033a8ebaf05a6a9dfd78c1cb/runtime/mangata-rococo/src/lib.rs#L66

https://github.com/mangata-finance/mangata-node/blob/25b2af6d95de3399033a8ebaf05a6a9dfd78c1cb/node/src/chain_spec/mangata_rococo.rs#L7

imstar15:
Update assets

image

@nvengal @chrisli30

runtime/turing/src/lib.rs
nvengal: This is unnecessary indirection/abstraction. These deps can be pulled directly from the originating xcm crate rather than the runtime file- https://github.com/OAK-Foundation/OAK-blockchain/blob/69eb46e21c06504497c340a0a391d00b7a53c1e3/node/src/chain_spec/turing.rs#L17-L20

nvengal: This is unnecessary indirection/abstraction. These deps can be pulled directly from the originating xcm crate rather than the runtime file- https://github.com/OAK-Foundation/OAK-blockchain/blob/69eb46e21c06504497c340a0a391d00b7a53c1e3/node/src/chain_spec/turing.rs#L17-L20

node/src/chain_spec/turing.rs
```diff
@@ -183,23 +183,73 @@ fn testnet_genesis(
...
+							},
+						},
+					),
+				),
+				(
+					3,
+					orml_asset_registry::AssetMetadata::<Balance, CustomMetadata>::encode(
+						&orml_asset_registry::AssetMetadata {
+							decimals: 18,
+							name: b"Shiden".to_vec(),
```
nvengal, line 240: There is no reason to have both `Rocstar` and `Shiden` in the development config. Only the one that is being used for development needs to be here.

node/src/chain_spec/turing.rs
```diff
@@ -183,23 +183,73 @@ fn testnet_genesis(
...
-					},
+					),
+				),
+				(
+					1,
+					orml_asset_registry::AssetMetadata::<Balance, CustomMetadata>::encode(
+						&orml_asset_registry::AssetMetadata {
+							decimals: 18,
+							name: b"Mangata".to_vec(),
+							symbol: b"MGX".to_vec(),
```
nvengal, line 209: `MGR` for mangata_rococo

node/src/chain_spec/turing.rs
```diff
@@ -183,23 +183,73 @@ fn testnet_genesis(
...
-					&orml_asset_registry::AssetMetadata {
-						decimals: TOKEN_DECIMALS,
-						name: "Native".as_bytes().to_vec(),
-						symbol: TOKEN_SYMBOL.as_bytes().to_vec(),
-						existential_deposit: 100_000_000,
-						location: Some(V1(MultiLocation { parents: 0, interior: Here })),
-						additional: CustomMetadata {
-							fee_per_second: Some(416_000_000_000),
-							conversion_rate: None,
+			assets: vec![
```
nvengal, line 186: See https://github.com/OAK-Foundation/OAK-blockchain/pull/319#discussion_r1048844541

The new assets should be moved out from this method.

nvengal (changes requested):

runtime/turing/src/lib.rs
imstar15: Ok, great idea! I will change it.

imstar15:
Code has been modified. Please review.

@nvengal @chrisli30 

Update assets
image

nvengal (approved):
Functionally looks fine. One suggestion for idiomatic rust.

node/src/chain_spec/turing.rs
```diff
@@ -126,11 +178,41 @@ fn testnet_genesis(
...
+		additional_assets,
+	]
+	.concat();
+
+	let mut last_asset_id = assets[0].0;
+	for asset in &assets {
+		if asset.0 > last_asset_id {
+			last_asset_id = asset.0;
+		}
+	}
```
nvengal, line 214:
```suggestion
let last_asset_id = assets.iter().map(|a| a.0).max().expect("atleast 1 item");
```

nvengal (approved):
Functionally, looks fine. One suggestion for more idiomatic rust.

chrisli30:
@imstar15, is this good to merge?

chrisli30 (approved):

chrisli30:
@imstar15, let’s update all existential value to 0.01 to be consistent with TUR. When you register an asset, remember to set it to 0.01.

This is what I’m registering on Turing Staging.
CleanShot 2022-12-29 at 10 47 18

chrisli30 (approved):

PR #320, opened 2022-12-16
Remove xcmp-handler fees rpc

nvengal:
Deprecated by `AutomationTime#queryFeeDetails`. Removing the endpoint from the node to avoid any confusion by partners on which fee rpc is the source of truth.

### DIFF
--- a/node/src/chain_spec/mod.rs
+++ b/node/src/chain_spec/mod.rs
@@ -19,6 +19,7 @@ pub mod oak;
 #[cfg(feature = "turing-node")]
 pub mod turing;
 
+#[allow(dead_code)]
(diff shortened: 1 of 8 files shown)


node/src/chain_spec/mod.rs
```diff
@@ -19,6 +19,7 @@ pub mod oak;
 #[cfg(feature = "turing-node")]
 pub mod turing;
 
+#[allow(dead_code)]
```
nvengal, line 22: Unrelated changes to clean up compiler warnings. This is not actually dead code, it is only present under some features. Rather than be pedantic about the features let's just suppress the warning.

chrisli30 (approved):
Yes, this is a good call. The changes look good to me.

@imstar15 FYI

imstar15 (approved):
looks good to me!

PR #321, opened 2022-12-27
Add support for XCM flow that begins with DescendOrigin

nvengal:
and pulls fees directly from the user account.

### DIFF
--- a/pallets/xcmp-handler/src/lib.rs
+++ b/pallets/xcmp-handler/src/lib.rs
@@ -44,11 +44,12 @@ pub mod migrations;
 pub mod weights;
 pub use weights::WeightInfo;
 
+use cumulus_primitives_core::ParaId;
+
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
 	use codec::Decode;
-	use cumulus_primitives_core::ParaId;
 	use frame_support::{
 		dispatch::DispatchResultWithPostInfo, pallet_prelude::*, traits::Currency,
 		weights::constants::WEIGHT_PER_SECOND,
@@ -107,6 +108,9 @@ pub mod pallet {
 
 		/// Utility for determining XCM instruction weights.
 		type Weigher: WeightBounds<<Self as pallet::Config>::Call>;
+
+		/// XCM Flow Selection Matcher
+		type XcmFlowSelector: XcmFlowSelector;
 	}
 
 	#[pallet::pallet]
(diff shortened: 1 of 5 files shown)


pallets/xcmp-handler/src/lib.rs
```diff
@@ -355,6 +375,59 @@ pub mod pallet {
...
+			transact_encoded_call_weight: u64,
+			xcm_weight: u64,
+			fee: u128,
+		) -> Result<
+			(xcm::latest::Xcm<<T as pallet::Config>::Call>, xcm::latest::Xcm<()>),
+			DispatchError,
+		> {
+			let target_asset_loc = T::CurrencyIdToMultiLocation::convert(currency_id)
+				.ok_or(Error::<T>::CurrencyNotSupported)?;
+			let target_asset =
```
nvengal, line 403: I'm not confident about this asset location specified in this way. We'll need to test it once more of the pieces are in place.

runtime/neumann/src/xcm_config.rs
```diff
@@ -166,11 +166,52 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+	}
+}
+
 pub type Barrier = DenyThenTry<
 	DenyReserveTransferToRelayChain,
 	(
 		TakeWeightCredit,
 		AllowTopLevelPaidExecutionFrom<Everything>,
+		// Allow xcm flow where execution fee is paid directly from DescendOrigin account
+		AllowPaidExecWithDescendOriginFrom<Everything>,
```
nvengal, line 214: Updates the barrier to allow xcm messages that start with DescendOrigin as described in Astar's docs- https://docs.astar.network/docs/xcm/building-with-xcm/xc-remote-transact#xcm-sequence

pallets/xcmp-handler/src/lib.rs
```diff
@@ -107,6 +111,9 @@ pub mod pallet {
 
 		/// Utility for determining XCM instruction weights.
 		type Weigher: WeightBounds<<Self as pallet::Config>::Call>;
+
+		/// XCM Flow Selection Matcher
+		type XcmFlowSelector: XcmFlowSelector;
```
nvengal, line 116: The idea with this design is that the runtime will pass in a matcher that decides which `ParaId` requires which xcm flow based on what the partner decides they want. This needs to be passed down from the runtime since the same `ParaId` might use two different flows in different envs.

pallets/xcmp-handler/src/lib.rs
```diff
@@ -570,3 +643,18 @@ impl<T: Config> XcmpTransactor<T::AccountId, T::CurrencyId> for Pallet<T> {
 		Ok(().into())
 	}
 }
+
+pub enum XcmFlow {
+	Normal,
+	Alternate,
```
nvengal, line 649: TODO: Alternate flow needs to bypass xcm fee collection

chrisli30, line 649: Let me ask around to see if there are more descriptive names.

chrisli30 (approved):
Looks good to me.

pallets/xcmp-handler/src/lib.rs
nvengal: Eventually we can have this be updatable by governance. In this alpha design it's set by the runtime developer.

PR #322, opened 2022-12-27
Add support for accepting foreign currency fees for automation scheduling (through xcm)

nvengal:
- Adds migration for shiden asset
- Updates automation-time fee handler to support foreign currencies

### DIFF
--- a/pallets/automation-time/src/fees.rs
+++ b/pallets/automation-time/src/fees.rs
@@ -61,8 +61,7 @@ where
 	) -> Result<R, DispatchError> {
 		let fee_handler = Self::new(owner, action, executions)?;
 		// Note: will need to account for fees in non-native tokens once we start accepting them
-		Self::can_pay_fee(owner, fee_handler.execution_fee.saturating_add(fee_handler.xcmp_fee))
-			.map_err(|_| Error::<T>::InsufficientBalance)?;
+		fee_handler.can_pay_fee().map_err(|_| Error::<T>::InsufficientBalance)?;
 		let outcome = prereq()?;
 		fee_handler.pay_fees()?;
 		Ok(outcome)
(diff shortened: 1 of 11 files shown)


pallets/automation-time/src/fees.rs
```diff
@@ -75,30 +76,36 @@ where
...
 		}
 
-		let withdraw_reason = WithdrawReasons::FEE;
-
-		match T::Currency::withdraw(who, fee, withdraw_reason, ExistenceRequirement::KeepAlive) {
-			Ok(imbalance) => {
-				OU::on_unbalanceds(Some(imbalance).into_iter());
+		match T::MultiCurrency::withdraw(self.currency_id, &self.owner, fee) {
+			Ok(_) => {
+				// TODO: fees need to go to multi currency enabled treasury
```
nvengal, line 108: This is a big `TODO`, but this pr was already getting big so I opted to separate this functionality. However, this is a blocker for release. Without implementing a treasury that can accept foreign currencies all automation fees will get burned.

nvengal: XcmFees in particular need to be held in a holding account, not burned.

runtime/turing/src/migrations/register_shiden_asset.rs
```diff
@@ -0,0 +1,119 @@
...
+			target: "asset_registry",
+			"on_runtime_upgrade: add shiden asset"
+		);
+
+		let asset = AssetMetadataOf {
+			decimals: 18,
+			name: b"Shiden".to_vec(),
+			symbol: b"SDN".to_vec(),
+			additional: CustomMetadata {
+				fee_per_second: Some(416_000_000_000),
```
nvengal, line 33: This is a placeholder. We need to decide on a real fee_per_sec before launch.

nvengal: fyi, @chrisli30

pallets/automation-time/src/fees.rs
chrisli30: Yes, noted and I agree.

runtime/turing/src/migrations/register_shiden_asset.rs
chrisli30: Let me ask the Astar team about the converstion rate, but let’s think about it, what is the best way to update its value? It doesn’t need to be changed frequently, for example just once a week. Could we write it from a whitelisted wallet address through a RPC?

chrisli30: The hard-coded value currently works for this PR and demo purpose.

nvengal: This value can be updated by governance. This brings up something I forgot to note- For turing-staging, we'll need to add the shibuya token with a governance proposal. Adding it here in a migration will also add it to turing since turing and turing-staging share one runtime.

chrisli30: okay, got it yeah as long as it’s adjustable with Governance. I’m registering the asset to Turing Staging.

chrisli30 (approved):
Currently no more questions on the PR.

PR #323, opened 2022-12-30
Bypass xcm fees for alternate flow

nvengal:
In the alternate flow fees are deducted directly from the user's derived account on the target chain. We don't need to hold or transfer xcm fees on our chain.


pallets/xcmp-handler/src/mock.rs
```diff
@@ -257,6 +260,17 @@ impl cumulus_pallet_xcm::Config for Test {
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 }
 
+pub struct XcmFlowMatcher;
+impl XcmFlowSelector for XcmFlowMatcher {
```
nvengal, line 264: Example usage for `XcmFlowSelector` that can be referenced if Shiden team decides they want to use the alternate flow. fyi, @chrisli30

chrisli30: This ALTERNATE_FLOW_PARA_ID should be a collection right? For example when both Mangata and Shiden choose the alternate flow.

Are our RPC and polkadot.js good at updating Array values? If not we will need to use a u128 to represent this array, by allocating every two bytes to each paraId.

For example, u128 will have 16 bytes(128/8=16) which is enough for 8 slots.
00_00_00_00_00_00_00_00

chrisli30: We need a more scalable way ...

nvengal: This var (`ALTERNATE_FLOW_PARA_ID`) is just being used for tests in this context so it's fine as a single value.

In a production context, the list of alternate flow para ids is not currently update-able without a code change. My thinking is that this is fine for now as I don't expect partners will want to change their flow often (or ever after launching). If we find that that's not the case we can move the map of alternate flow para ids to runtime storage and expose it through governance (note: rpcs are read-only).

I didn't implement the more complex approach for this upfront because users pay a cost penalty for each new storage read. Moving this value to storage would mean an additional read for every scheduling and execution of an automation. Not a huge deal, but should be avoided if it's unnecessary.

chrisli30: Nvm, just realize this is a test code to turn on Alternate flow for a single chain.

chrisli30 (approved):
Looks good to me.

pallets/xcmp-handler/src/mock.rs
chrisli30: @nvengal 

Yeah, you are right. The flow path is choosen by parachain on init, but not by individual users.

I agree. Let’s minimize storage read cost and keep it as-is.

PR #324, opened 2022-12-30
Make asset_registry migration idempotent

nvengal:
Probably unnecessary, but better safe than sorry since it might be some time before this gets deployed.


chrisli30 (approved):
LGTM

PR #330, opened 2023-01-30
Update frame-weight-template for Weight v2

nvengal:


chrisli30 (approved):
Got it. Looks good to me.

PR #334, opened 2023-02-14
Use number of xcm instructions for weight calculations

nvengal:


pallets/xcmp-handler/src/lib.rs
```diff
@@ -239,8 +239,12 @@ pub mod pallet {
 		) -> Result<(u128, u64), DispatchError> {
 			let xcm_data = XcmChainCurrencyData::<T>::get(para_id, currency_id)
 				.ok_or(Error::<T>::CurrencyChainComboNotFound)?;
+
+			let (_, target_instructions) = Self::xcm_instruction_skeleton(para_id, currency_id)?;
 			let weight = xcm_data
 				.instruction_weight
+				.checked_mul(target_instructions.len() as u64)
+				.ok_or(Error::<T>::WeightOverflow)?
```
chrisli30, line 247: Is this `.ok_or` check here to throw early and skip the math of the next line?

xcm weight is tiny compared to that of transact, so if we want to add a check to the middle it’s best to apply it to `transact_encoded_call_weight` first.

pallets/xcmp-handler/src/mock.rs
```diff
@@ -284,7 +292,7 @@ impl pallet_xcmp_handler::Config for Test {
 	type GetNativeCurrencyId = GetNativeCurrencyId;
 	type SelfParaId = parachain_info::Pallet<Test>;
 	type AccountIdToMultiLocation = AccountIdToMultiLocation;
-	type CurrencyIdToMultiLocation = ();
+	type CurrencyIdToMultiLocation = TokenIdConvert;
```
chrisli30, line 295: Is this `CurrencyIdToMultiLocation` tested?

In other words, is there a test passing with () before and with TokenIdConvert now, or there’s none for it.

chrisli30 (approved):
A couple of comments, but non-blocking. I’m approving this PR.

pallets/xcmp-handler/src/mock.rs
nvengal: Good question. Before the changes in this PR the CurrencyIdToMultiLocation was not being adequately tested by the existing tests. This pr changed that by generating the xcm skeleton for fees. With the new fees code we need to use are real mock for TokenIdConvert.

If this change is removed, this is the failing test:
```
---- tests::calculate_xcm_fee_doesnt_charge_for_alternate_flow stdout ----
thread 'tests::calculate_xcm_fee_doesnt_charge_for_alternate_flow' panicked at 'assertion failed: `(left == right)`
  left: `Err(Module(ModuleError { index: 3, error: [9, 0, 0, 0], message: Some("CurrencyNotSupported") }))`,
 right: `Ok((0, 700000000))`', pallets/xcmp-handler/src/tests.rs:228:9
```

pallets/xcmp-handler/src/lib.rs
nvengal: Technically, you're correct but the reality is that this will never fail under normal runtime conditions. To fail the weight would need to overflow the u64 max `18_446_744_073_709_551_615`.
The `ok_or` pattern here is a rust idiom for unwrapping the result type. `checked_mul` returns a `Result<u64, Error>` so we need to unwrap the result to get the underlying value. Also, to preserve the correct order of operations we need to do the multiplication before (or separately) from the addition.

chrisli30: Gotcha, the unwrap is necessary, so let’s keep it as-is.

pallets/xcmp-handler/src/mock.rs
chrisli30: Cool, so the conversion is covered now.

nvengal: yup 👍

chrisli30 (approved):

PR #336, opened 2023-02-15
Remove old migrations from runtime executor

nvengal:
This is a required step after every release to prevent migrations from being re-run on the next release


chrisli30 (approved):
Got it, that makes sense 👍

PR #337, opened 2023-02-17
Add dynamic xcm flow config

nvengal:


pallets/xcmp-handler/src/migrations/add_flow_to_currency_data.rs
```diff
@@ -0,0 +1,91 @@
...
+impl From<OldXcmCurrencyData> for XcmCurrencyData {
+	fn from(data: OldXcmCurrencyData) -> Self {
+		XcmCurrencyData {
+			native: data.native,
+			fee_per_second: data.fee_per_second,
+			instruction_weight: data.instruction_weight,
+			flow: XcmFlow::Normal,
+		}
+	}
+}
```
chrisli30, line 34: I see. The flow is now defined in foreign currency data. Ideally, we should detach this chain-level config from currency, but it is a good place and maybe the only place for now.

chrisli30 (commented):
This PR is quite some refactor. 👍  Let’s go over it Friday morning.

pallets/xcmp-handler/src/migrations/add_flow_to_currency_data.rs
chrisli30: No additional storage read cost within one transaction in `xcm-handler` metadata.

chrisli30 (approved):

node/src/chain_spec/turing.rs
```diff
@@ -88,13 +89,29 @@ pub fn turing_development_config() -> ChainSpec {
...
 						false,
 						5_376_000_000_000,
 						1_000_000_000,
+						XcmFlow::Normal,
+					),
+					(
+						2000,
+						turing_runtime::NATIVE_TOKEN_ID,
+						false,
+						21_250_000_000_000_000_000_000,
```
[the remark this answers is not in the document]
chrisli30: It’s an esitmate through testing, which should be higher than the actual fee and work.

PR #341, opened 2023-02-23
🔧 lower minCandidateStk

SimonKraus:
As discussed with Dawufi within ChaosDAO discord the current configuration of a 2 million minimum candidate stake on Turing network does not work out well. 
The average total backing per collator in Turing is currently around 1.25M TUR so a node joining with 2M TUR would earn less than delegating.
There currently is a vacant collator slot because Nova Wallet team left.

ChaosDAO wants to support Turing and operate a collator node for the network.

Therefore we proposed to lower the minCandidateStk to 1M TUR and created this PR to save you the work and get it applied.

### DIFF
--- a/node/src/chain_spec/neumann.rs
+++ b/node/src/chain_spec/neumann.rs
@@ -12,6 +12,7 @@ use crate::chain_spec::{
 	get_account_id_from_seed, get_collator_keys_from_seed, inflation_config, DummyChainSpec,
 	Extensions,
 };
+use codec::Encode;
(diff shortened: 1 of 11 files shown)


chrisli30:
@SimonKraus, I agree. The threshold of 2M was set last year but the value is not really applicable based on the current average staked amount. 

This code change looks no problem, but in order to update that value on-chain we will need to start a governance proposal.

chrisli30 (approved):

PR #345, opened 2023-03-02
Fix Moonbase multilocation

imstar15:


node/src/chain_spec/turing.rs
```diff
@@ -16,9 +16,9 @@ use turing_runtime::{
 	VestingConfig, XcmpHandlerConfig,
 };
 use xcm::{
-	latest::prelude::X1,
+	latest::prelude::{X1, X2},
 	opaque::latest::{Junctions::Here, MultiLocation},
-	v1::Junction::Parachain,
+	v1::Junction::{PalletInstance, Parachain},
```
chrisli30, line 21: This changes the global object structure of xcm? Will it affect the existing xcm calls such as those with Mangata and Shiden?

imstar15: No, it's only related to Moonbase.

chrisli30 (approved):

PR #350, opened 2023-04-13
Bump to spec version 290

imstar15:


node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "oak-node"
-version = "1.8.0"
+version = "1.8.2"
```
chrisli30, line 3: Do not change node version if there’s no change to the node (rpc endpoints)

Cargo.lock
```diff
@@ -4879,7 +4879,7 @@ dependencies = [
 
 [[package]]
 name = "oak-node"
-version = "1.8.0"
```
chrisli30, line 4882: Do not change node version if there’s no change to the node (rpc endpoints)

chrisli30 (changes requested):

chrisli30 (approved):

PR #352, opened 2023-04-20
Fix AutoCompoundDelegatedStake task

imstar15:


pallets/automation-time/src/migrations/upgrade_xcmp_task_struct.rs
```diff
@@ -51,7 +51,6 @@ pub enum OldAction<T: Config> {
 		delegator: T::AccountId,
 		collator: T::AccountId,
 		account_minimum: BalanceOf<T>,
-		frequency: Seconds,
```
chrisli30, line 54: Would this change the interface(parameters) of the existing `scheduleAutoCompoundDelegatedStakeTask` call on Turing Network?

In other words, what is the parameter list of `scheduleAutoCompoundDelegatedStakeTask` after this migration?

chrisli30 (changes requested):

pallets/automation-time/src/migrations/upgrade_xcmp_task_struct.rs
chrisli30: This line is not needed because the Schedule data is already migrated to Fixed and Recurring formats on Turing Network

chrisli30 (approved):
LGTM. The issue was due to old migration cold not deleted after [v1.8.0](https://github.com/OAK-Foundation/OAK-blockchain/releases/tag/v1.8.0)

PR #361, opened 2023-05-23
Add pallet-multisig/std to Cargo.toml in runtime

imstar15:
Fix compliant error with srtool


chrisli30 (approved):
Discussed this with Charles. std libraries could work.

PR #363, opened 2023-05-30
Add sudo feature to dev environment

imstar15:


chrisli30 (changes requested):
I am not sure if we should add this feature before the separation of Turing and Staging code. Sudo should only be added to Turing Staging but not Turing. Let’s discuss it and leave this PR as a draft for now.

chrisli30:
Re-open this ticket for discussion.

PR #365, opened 2023-06-06
Add schedule_as into TaskScheduled event

imstar15:
Fixed the problem that when calling `scheduleXcmpTaskThroughProxy`, the `who` parameter of TaskScheduled was inconsistent with `AccountId32` in `automationTime.accountTasks` in storage.

`who` is always the caller for extrinsic.

image

image

### DIFF
--- a/pallets/automation-time/src/benchmarking.rs
+++ b/pallets/automation-time/src/benchmarking.rs
@@ -286,7 +286,7 @@ benchmarks! {
 		let task_id = Pallet::<T>::generate_task_id(caller.clone(), provided_id.clone());
 	}: schedule_dynamic_dispatch_task(RawOrigin::Signed(caller.clone()), provided_id, schedule, Box::new(call))
 	verify {
-		assert_last_event::<T>(Event::TaskScheduled { who: caller, task_id: task_id }.into())
+		assert_last_event::<T>(Event::TaskScheduled { who: caller, task_id: task_id, schedule_as: None }.into())
 	}
 
 	schedule_dynamic_dispatch_task_full {
@@ -312,7 +312,7 @@ benchmarks! {
 		schedule_notify_tasks::<T>(caller.clone(), times, T::MaxTasksPerSlot::get() - 1);
 	}: schedule_dynamic_dispatch_task(RawOrigin::Signed(caller.clone()), provided_id, schedule, Box::new(call))
 	verify {
-		assert_last_event::<T>(Event::TaskScheduled { who: caller, task_id: task_id }.into())
+		assert_last_event::<T>(Event::TaskScheduled { who: caller, task_id: task_id, schedule_as: None }.into())
 	}
 
 	cancel_scheduled_task_full {
--- a/pallets/automation-time/src/mock.rs
+++ b/pallets/automation-time/src/mock.rs
@@ -351,7 +351,11 @@ impl Convert<CurrencyId, Option<MultiLocation>> for MockTokenIdConvert {
 pub struct MockEnsureProxy;
 impl EnsureProxy<AccountId> for MockEnsureProxy {
 	fn ensure_ok(_delegator: AccountId, _delegatee: AccountId) -> Result<(), &'static str> {
-		Ok(())
+		if _delegator == ALICE.into() && _delegatee == BOB.into() {
+			Ok(())
+		} else {
+			Err("proxy error: expected `ProxyType::Any`")
+		}
 	}
 }
 
--- a/pallets/automation-time/src/mock.rs
+++ b/pallets/automation-time/src/mock.rs
@@ -42,6 +42,9 @@ pub type CurrencyId = u32;
 
 pub const ALICE: [u8; 32] = [1u8; 32];
 pub const BOB: [u8; 32] = [2u8; 32];
+pub const DELEGATOR_ACCOUNT: [u8; 32] = [3u8; 32];
+pub const PROXY_ACCOUNT: [u8; 32] = [4u8; 32];
+
 pub const PARA_ID: u32 = 2000;
 pub const NATIVE: CurrencyId = 0;
 
@@ -351,7 +354,7 @@ impl Convert<CurrencyId, Option<MultiLocation>> for MockTokenIdConvert {
 pub struct MockEnsureProxy;
 impl EnsureProxy<AccountId> for MockEnsureProxy {
 	fn ensure_ok(_delegator: AccountId, _delegatee: AccountId) -> Result<(), &'static str> {
-		if _delegator == ALICE.into() && _delegatee == BOB.into() {
+		if _delegator == DELEGATOR_ACCOUNT.into() && _delegatee == PROXY_ACCOUNT.into() {
 			Ok(())
 		} else {
 			Err("proxy error: expected `ProxyType::Any`")
--- a/pallets/automation-time/src/tests.rs
+++ b/pallets/automation-time/src/tests.rs
@@ -792,7 +792,8 @@ mod extrinsics {
 					last_event(),
 					RuntimeEvent::AutomationTime(crate::Event::TaskScheduled {
 						who: account_id,
-						task_id
+						task_id,
+						schedule_as: None,
 					})
 				);
 			})
--- a/pallets/automation-time/src/tests.rs
+++ b/pallets/automation-time/src/tests.rs
@@ -350,7 +350,7 @@ fn schedule_xcmp_through_proxy_works() {
 				}) if *who == proxy_account && *schedule_as == Some(delegator_account.clone()) => true,
 				_ => false,
 			})
-			.expect("The TaskScheduled event should be emitted, with the data where who is the proxy_account, and schedule_as is the user_account.");
+			.expect("TaskScheduled event should emit with 'who' being proxy_account, and 'schedule_as' being delegator_account.");
 	})
 }
 
--- a/pallets/automation-time/src/tests.rs
+++ b/pallets/automation-time/src/tests.rs
@@ -271,8 +271,8 @@ fn schedule_xcmp_works() {
 fn schedule_xcmp_through_proxy_works() {
 	new_test_ext(START_BLOCK_TIME).execute_with(|| {
 		let provided_id = vec![50];
-		let user_account = AccountId32::new(ALICE);
-		let proxy_account = AccountId32::new(BOB);
+		let delegator_account = AccountId32::new(DELEGATOR_ACCOUNT);
+		let proxy_account = AccountId32::new(PROXY_ACCOUNT);
 		let call: Vec<u8> = vec![2, 4, 5];
 
 		// Funds including XCM fees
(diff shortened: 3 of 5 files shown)


chrisli30 (changes requested):
I’m not confident to check this in without unit tests. When you get time, please write down **the design** of unit test cases to make sure event filing for this xcmp task is right with a variety of inputs.

imstar15:
The design of unit test cases:
1. The proxy account Bob invokes scheduleXcmpTaskThroughProxy with scheduleAs: Alice.
2. In the emitted TaskScheduled event data, who should be Bob, and scheduleAs should be Alice.
3. In the accountTasks storage, it can be queried that the owner of this task is Bob.

@chrisli30

pallets/automation-time/src/tests.rs
```diff
@@ -267,6 +267,49 @@ fn schedule_xcmp_works() {
 	})
 }
 
+#[test]
+fn schedule_xcmp_through_proxy_works() {
+	new_test_ext(START_BLOCK_TIME).execute_with(|| {
+		let provided_id = vec![50];
+		let alice = AccountId32::new(ALICE);
+		let bob = AccountId32::new(BOB);
```
chrisli30, line 275: Do not use alice or bob as variable, since they have no functional meaning. If bob is the signer, name the wallet owner or signer.

pallets/automation-time/src/tests.rs
```diff
@@ -267,6 +267,49 @@ fn schedule_xcmp_works() {
...
+			alice.clone(),
+		));
+
+		let tasks = AutomationTime::get_scheduled_tasks(SCHEDULED_TIME);
+		assert_eq!(tasks.is_some(), true);
+
+		let tasks = tasks.unwrap();
+		assert_eq!(tasks.tasks[0].0, bob.clone());
+
+		events()
```
chrisli30, line 299: Add comments to this long events() block, as in what is being verified.

pallets/automation-time/src/tests.rs
```diff
@@ -267,6 +267,49 @@ fn schedule_xcmp_works() {
...
+			.into_iter()
+			.find(|e| match e {
+				RuntimeEvent::AutomationTime(crate::Event::TaskScheduled {
+					who,
+					schedule_as,
+					..
+				}) if *who == bob && *schedule_as == Some(alice.clone()) => true,
+				_ => false,
+			})
+			.expect("TaskScheduled event should have been emitted with correct parameters");
```
chrisli30, line 309: Update this error information. The current line is unclear.

According to you, this PR is to "Fixed the problem that when calling scheduleXcmpTaskThroughProxy, the who parameter of TaskScheduled was inconsistent with AccountId32 in automationTime.accountTasks in storage."

pallets/automation-time/src/tests.rs
```diff
@@ -267,6 +267,49 @@ fn schedule_xcmp_works() {
...
+			RuntimeOrigin::signed(bob.clone()),
+			provided_id,
+			ScheduleParam::Fixed { execution_times: vec![SCHEDULED_TIME] },
+			PARA_ID.try_into().unwrap(),
+			NATIVE,
+			MultiLocation::new(1, X1(Parachain(PARA_ID.into()))).into(),
+			call.clone(),
+			Weight::from_ref_time(100_000),
+			alice.clone(),
+		));
```
chrisli30, line 291: There’s no connection between the wallets alice and bob. We are not checking if one is the other’s proxy wallet, correct?

pallets/automation-time/src/tests.rs
```diff
@@ -267,6 +267,49 @@ fn schedule_xcmp_works() {
 	})
 }
 
+#[test]
```
chrisli30, line 270: We need to include more tests around this function. For example,
1. schedule through an empty wallet address(empty string).
2. schedule through an invalid Turing wallet address.
3. schedule through a wallet address that is the same as the owner wallet.
4. schedule with an invalid multi-location
5. schedule with a non-existent multi-location
6. Could verify that the task can execute with a correct proxy?

chrisli30 (changes requested):

pallets/automation-time/src/benchmarking.rs
```diff
@@ -312,7 +312,7 @@ benchmarks! {
 		schedule_notify_tasks::<T>(caller.clone(), times, T::MaxTasksPerSlot::get() - 1);
 	}: schedule_dynamic_dispatch_task(RawOrigin::Signed(caller.clone()), provided_id, schedule, Box::new(call))
 	verify {
-		assert_last_event::<T>(Event::TaskScheduled { who: caller, task_id: task_id }.into())
+		assert_last_event::<T>(Event::TaskScheduled { who: caller, task_id: task_id, schedule_as: None }.into())
```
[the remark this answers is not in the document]
imstar15: For any dynamic dispatch task, it should generate a `TaskScheduled` event where `schedule_as` is `None`.

In the `verify` code block, assert the `schedule_as` in the `TaskScheduled` event is `None`, which is correct.

We cannot construct a dynamic dispatch task where schedule_as has a real value.

pallets/automation-time/src/tests.rs

pallets/automation-time/src/tests.rs
imstar15: Fixed

pallets/automation-time/src/tests.rs

pallets/automation-time/src/tests.rs
imstar15: This is the implementation of `MockEnsureProxy`, which is only used for automation-time testing purposes.

https://github.com/OAK-Foundation/OAK-blockchain/blob/e8415385756db9daf29cf12cb2ae49e2772d4c7d/pallets/automation-time/src/mock.rs#L352

pallets/automation-time/src/tests.rs
imstar15: **1. schedule through an empty wallet address(empty string).**
The AccountId32::new(bytes) method requires a byte array parameter with 32 elements. An empty wallet address cannot be used to construct an AccountId32 variable.

**2. schedule through an invalid Turing wallet address.**
The code within the automation time pallet does not perform validation checks on the account addresses.

Account address validation is a feature provided by the Substrate framework, which verifies the correctness of user addresses during actual runtime. 

However, in functional testing, incorrect account addresses will not result in an error.

pallets/automation-time/src/tests.rs
chrisli30: Okay, so these two accounts are linked. This question is resolved.
```
if _delegator == ALICE.into() && _delegatee == BOB.into() {
			Ok(())
		}
```

pallets/automation-time/src/benchmarking.rs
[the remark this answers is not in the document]
chrisli30: Okay, got it. Yeah dynamic dispatch task should use the value None. 👍

pallets/automation-time/src/mock.rs
```diff
@@ -351,7 +351,11 @@ impl Convert<CurrencyId, Option<MultiLocation>> for MockTokenIdConvert {
 pub struct MockEnsureProxy;
 impl EnsureProxy<AccountId> for MockEnsureProxy {
 	fn ensure_ok(_delegator: AccountId, _delegatee: AccountId) -> Result<(), &'static str> {
-		Ok(())
+		if _delegator == ALICE.into() && _delegatee == BOB.into() {
```
chrisli30, line 354: If this ALICE is not used anywhere else we should call it `OWNER_ACCOUNT` and BOB for `PROXY_ACCOUNT`, where OWNER_ACCOUNT is the delegator and PROXY_ACCOUNT is a delegatee.

pallets/automation-time/src/tests.rs
```diff
@@ -271,42 +271,70 @@ fn schedule_xcmp_works() {
...
 				RuntimeEvent::AutomationTime(crate::Event::TaskScheduled {
 					who,
 					schedule_as,
 					..
-				}) if *who == bob && *schedule_as == Some(alice.clone()) => true,
+				}) if *who == proxy_account && *schedule_as == Some(user_account.clone()) => true,
 				_ => false,
 			})
-			.expect("TaskScheduled event should have been emitted with correct parameters");
+			.expect("The TaskScheduled event should be emitted, with the data where who is the proxy_account, and schedule_as is the user_account.");
```
chrisli30, line 310: TaskScheduled event should emit with 'who' being proxy_account, and 'schedule_as' being owner_account.

Do you know how to insert the real value of proxy_account into this `expect()` string? The word 'proxy_account' doesn’t contain any actual information is this error is thrown.

pallets/automation-time/src/tests.rs
```diff
@@ -271,42 +271,70 @@ fn schedule_xcmp_works() {
...
+				provided_id,
+				ScheduleParam::Fixed { execution_times: vec![SCHEDULED_TIME] },
+				PARA_ID.try_into().unwrap(),
+				NATIVE,
+				MultiLocation::new(1, X1(Parachain(PARA_ID.into()))).into(),
+				call.clone(),
+				Weight::from_ref_time(100_000),
+				user_account.clone(),
+			),
+			sp_runtime::DispatchError::Other("proxy error: expected `ProxyType::Any`"),
```
chrisli30, line 336: To double check, does the proxy error thrown during task schedule, or upon task execution later?

chrisli30 (changes requested):

pallets/automation-time/src/tests.rs
chrisli30: The answer to the above question is yes.

pallets/automation-time/src/tests.rs
chrisli30: 5. a wrong `xcm_asset_location` for destination doesn’t cause error on our XcmMessageSent, but causing one on relay chain instead. Therefore, we can’t test the validity in our chain’s code.

pallets/automation-time/src/tests.rs
```diff
@@ -271,42 +271,43 @@ fn schedule_xcmp_works() {
 fn schedule_xcmp_through_proxy_works() {
 	new_test_ext(START_BLOCK_TIME).execute_with(|| {
 		let provided_id = vec![50];
-		let alice = AccountId32::new(ALICE);
-		let bob = AccountId32::new(BOB);
+		let user_account = AccountId32::new(ALICE);
```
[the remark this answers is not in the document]
chrisli30: We decide to call them delegator_account and proxy_account.

pallets/automation-time/src/mock.rs
```diff
@@ -351,7 +351,11 @@ impl Convert<CurrencyId, Option<MultiLocation>> for MockTokenIdConvert {
...
+			Ok(())
+		} else {
+			Err("proxy error: expected `ProxyType::Any`")
+		}
```
chrisli30, line 358: Let’s add another set accounts for proxy related tests.

chrisli30 (approved):
Looks good to me!

PR #367, opened 2023-06-13
Unblock reserved transfer of asset on relay chain to self

imstar15:


chrisli30 (approved):
The denying relay chain token transfer was an issue in early 2022, but no longer required in our code. This looks good.

PR #368, opened 2023-06-13
Display polkadotXcm storage data in chain state -> storage

imstar15:
image


chrisli30 (approved):
Both Moonriver and Shiden chose to display the storage data of the polkadotXcm pallet, so we are doing the same.

chrisli30:
@imstar15 the build release github action failed. Would re-run fix it, or there is some issues with this code change?

PR #369, opened 2023-06-14
Remove migration in runtime 293

imstar15:


runtime/neumann/src/xcm_config.rs
```diff
@@ -4,9 +4,9 @@ use super::{
 	UnknownTokens, XcmpQueue, MAXIMUM_BLOCK_WEIGHT, NATIVE_TOKEN_ID,
 };
 
-use core::marker::PhantomData;
+// use core::marker::PhantomData;
```
chrisli30, line 7: Remove this line as well?

chrisli30 (approved):
Looks good. There’s only one comment.

runtime/neumann/src/xcm_config.rs
imstar15: Yes, fixed.

PR #373, opened 2023-06-16
deprecate schedule_notify_task

v9n:
This PR remove `schedule_notify_task` extrinsic. This extrinsic is also used to setup and run test for many scenario so i migrated them to dynamic dispatch call and assert their behaviour.

### DIFF
--- a/pallets/automation-time/src/tests.rs
+++ b/pallets/automation-time/src/tests.rs
@@ -577,19 +577,19 @@ fn schedule_time_slot_full() {
 	})
 }
 
-// test case when a lot in full, we will roll back its state atomically
-// and won't leave the tasks queue in a partial state.
+// test case when a slot in full, we will roll back its state atomically
+// and won't leave the task queue in a partial state.
 //
-// It's similar to above test. However, we test the tasks that has scheduled
-// with many execution_times where as only a field execution_time slot is full
-// but the rest of execution_time slots aren't full.
+// It's similar to above test. However, we test a task that has scheduled
+// with many execution_times where as only a few execution_time slots are full
+// while the rest of execution_time slots aren't full.
 //
-// even though other time slot has not full, we still reject, return TimeSlotFull
-// error and verify  that none of the tasks has been scheduled into any
+// even though other time slots aren't full, we still reject as a whole, return
+// TimeSlotFull error and verify that none of the tasks has been scheduled into any
 // time slot, even the one that isn't full.
 //
-// in other word, task scheduled is atomic, all task execution needs to be able
-// to put into the schedule tasks slot, otherwise none of data should be stored
+// in other word, task scheduled is atomic, all task executions need to be able
+// to put into the schedule task slots, otherwise none of data should be stored
 // partially
 #[test]
 fn schedule_time_slot_full_rolls_back() {


pallets/automation-time/src/lib.rs
```diff
@@ -1050,8 +1010,17 @@ pub mod pallet {
...
 		/// Fire the notify event with the custom message.
-		pub fn run_notify_task(message: Vec<u8>) -> (Weight, Option<DispatchError>) {
-			Self::deposit_event(Event::Notify { message });
+		pub fn run_notify_task(message: Vec<u8>, block_number: u32) -> (Weight, Option<DispatchError>) {
+            // when re-execution from the old block, we will continue to emit the even normally
+            // and only cut off support at this block height
+            // the extrinsic is disabled so no one can schedule that kind of tasks, just old tasks
+            // TODO: Find out turing staging number and initalize accordingly
+            // This is the last block we  have this event
+            let target_block: u32 = 1244899;
```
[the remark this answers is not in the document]
chrisli30: Good question. The blockchain itself will run data migration right before new runtime code during an upgrade, so the data needs to be correctly transformed. In order to achieve that, as long as there’s data change in storage, migration could needs to follow.

We write the migration code and use [try-runtime](https://github.com/OAK-Foundation/OAK-blockchain/blob/master/scripts/try-runtime.sh) script to verify the correctness.

After the release of the code, those one-time data migration needs to be removed so they don’t run in the next release. Here’s the most recent example, https://github.com/OAK-Foundation/OAK-blockchain/pull/369

pallets/automation-time/src/tests.rs
```diff
@@ -90,90 +90,6 @@ fn schedule_too_far_out() {
...
-				vec![SCHEDULED_TIME],
-				vec![]
-			),
-			Error::<Test>::EmptyMessage,
-		);
-	})
-}
-
-#[test]
-fn schedule_no_provided_id() {
```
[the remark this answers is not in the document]
chrisli30: Yes, I agree. All three tests should be rewritten.

pallets/automation-time/src/lib.rs
```diff
@@ -349,63 +349,21 @@ pub mod pallet {
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
-		fn on_initialize(_: T::BlockNumber) -> Weight {
+		fn on_initialize(block: T::BlockNumber) -> Weight {
```
chrisli30, line 352: If the block variable is not used in this function, maybe we should call it `_block`? ( `_` is definitely not meaningful and should not be used.)

From ChatGPT:
In Rust, using an underscore (`_`) as a variable or parameter name is a convention to indicate that the value is intentionally unused or ignored. It is used when a variable or parameter is required by the language syntax but not actually needed within the code block.

In the specific example you provided (`fn on_initialize(block: T::BlockNumber) -> Weight`), if the `block` parameter is not used within the body of the `on_initialize` function, an underscore (`_`) could be used instead to indicate that the parameter is intentionally ignored. This can be helpful for code readability and to avoid compiler warnings about unused variables.

Here's an example of how the code could be written using `_` to indicate the parameter is unused:

```rust
fn on_initialize(_block: T::BlockNumber) -> Weight {
    // Code logic here
    // The `_block` parameter is intentionally ignored
}
```

Using `_` as a variable or parameter name is not specific to the `Block` type; it can be applied to any variable or parameter in Rust.

pallets/automation-time/src/mock.rs
```diff
@@ -389,19 +389,26 @@ pub fn new_test_ext(state_block_time: u64) -> sp_io::TestExternalities {
 	ext
 }
 
+// A function to support test scheduleing.
+// We don't focus on making sure the execution run properly. We just focus on
+// making sure a task is scheduled into the queue
```
chrisli30, line 394: Love the comment!

pallets/automation-time/src/tests.rs
```diff
@@ -38,12 +38,16 @@ const LAST_BLOCK_TIME: u64 = START_BLOCK_TIME / 1_000;
 #[test]
 fn schedule_invalid_time() {
```
chrisli30, line 39: Since all timeAutomation call has two kinds of scheduling, Fixed and Recurring, we should duplicate **time related tests** and **test both types**.

For Fixed, multiple timestamps need always to be tested since the input is an array, Vec<UnixTime>
For Recurring, only one timestamp needs to be tested, but there’s a second parameter `frequency` to test as well, `{ next_execution_time: UnixTime, frequency: Seconds}`.

pallets/automation-time/src/tests.rs
```diff
@@ -476,16 +428,29 @@ fn schedule_time_slot_full() {
 #[test]
 fn schedule_time_slot_full_rolls_back() {
```
chrisli30, line 429: Could you comment this roll back test case to explain its purpose?

pallets/automation-time/src/tests.rs
```diff
@@ -447,26 +392,33 @@ fn schedule_execution_times_removes_dupes() {
 #[test]
 fn schedule_time_slot_full() {
```
chrisli30, line 393: Could you comment this slot full test case to explain its purpose?

pallets/automation-time/src/tests.rs
```diff
@@ -990,11 +969,15 @@ fn trigger_tasks_handles_missed_slots() {
 #[test]
 fn trigger_tasks_limits_missed_slots() {
```
chrisli30, line 970: Could you comment this missed slots test case to explain its purpose?

pallets/automation-time/src/types.rs
```diff
@@ -442,8 +448,17 @@ mod tests {
 				scheduled_tasks
```
chrisli30, line 448: Could you comment this try_push_works_when_slot_is_not_full test case as well to explain its purpose?

pallets/automation-time/src/types.rs
```diff
@@ -220,6 +222,8 @@ impl<AccountId: Ord, Balance: Ord, CurrencyId: Ord> PartialEq
 
 impl<AccountId: Ord, Balance: Ord, CurrencyId: Ord> Eq for Task<AccountId, Balance, CurrencyId> {}
 
+use sp_runtime::print;
```
chrisli30, line 225: We should remove this line

pallets/automation-time/src/tests.rs
```diff
@@ -619,9 +584,23 @@ fn cancel_works_for_an_executed_task() {
 		}
 
 		AutomationTime::trigger_tasks(Weight::from_ref_time(200_000));
+		let my_events = events();
+
+		println!("event {:?}", my_events);
```
chrisli30, line 589: Temporary console println should all be removed in code PR stage.

chrisli30 (changes requested):
@v9n I have requested to comment the purpose of multiple tests which require explanation. Reading through those test code and figuring out their purpose will be helpful in understanding the time automation implementation.

chrisli30:
@imstar15 could you review this PR as well?

pallets/automation-time/src/lib.rs
imstar15: Yes, please call it `_block`.

pallets/automation-time/src/tests.rs
```diff
@@ -78,102 +82,19 @@ fn schedule_past_time() {
...
-				vec![SCHEDULED_TIME],
-				vec![12]
-			),
-			Error::<Test>::EmptyProvidedId,
-		);
-	})
-}
-
-#[test]
-fn schedule_not_enough_for_fees() {
```
imstar15, line 124: I believe that `schedule_not_enough_for_fees` should be retained. Please use the `schedule_dynamic_dispatch_task` call to implement this test.

imstar15 (changes requested):

pallets/automation-time/src/tests.rs
chrisli30: Yes, I agree. schedule_not_enough_for_fees should be rewritten. Let’s discuss it Vinh

chrisli30 (approved):
LGTM

imstar15 (approved):
Looks good to me!

vlpatel:
The number of the PR that added democracy to OAK

PR #380, opened 2023-07-05
Remove schedule notify

v9n:
This PR is exactly the same as https://github.com/OAK-Foundation/OAK-blockchain/pull/373 which is merged but because we don't want stacking PRs so I re-do them again to merge to the main branch as it's instead of stacking multiple PRs

Proof of commit hash

Screenshot 2023-07-05 at 9 09 00 AM

other pr(which is approved, and having the same code as this one)

Screenshot 2023-07-05 at 9 09 17 AM


chrisli30 (approved):
Yes, this one is good to merge.

PR #381, opened 2023-07-06
auto lint code with `cargo fmt`

v9n:
This is just a utility github action to auto run `cargo fmt` and if detect changes due to auto format, it will commit the change to the same branch, thus avoid us need to run this manually or having setup a git hooks locally to catch this.

it helps to simplify the code because the repeated work about re-runing `cargo fmt` can be outsource to the CI without relying on a local editor setup to auto run, or manual run.'

It looks like this in practice. The `fixup:...` one is the auto format commit run in this action
Screenshot 2023-07-06 at 12 03 16 AM


chrisli30 (approved):
Yep, it looks good. Let’s set it up.

PR #383, opened 2023-07-13
Fix #387 - Generalize the Event and Error of tasks by separating execution from triggering and rescheduling

### LINKED ISSUE #387: [OIP] Solve TaskScheduled and TaskRescheduled duplicate problem
## OAK Improvement Proposal

**Motivation**

Looking at the task history on Turing Network, specifically the task with the hash 0xcae095bc2b84249e4ca65671759808e0bb10680fb4fedd1ad9e358e4e084de39 shown in the screenshot below, it is evident that after each execution, both TaskScheduled and TaskRescheduled events contain identical data.

CleanShot 2023-07-20 at 17 57 54

Another issue is that the event of task triggering is intertwined with execution and has no indication of whether to reschedule for the next execution. To address this, I propose the following standard for task events.

For all tasks,
After the creation extrinsic, a `TaskScheduled` needs to be fired. ✅ 

During task execution, 
1. A `TaskTriggered` event should be fired to indicate the successful triggering of a task.
Event Name: TaskTriggered
Attributes: 
    - taskId:
    - who:
    - condition: { type:`time`, value: `1689903366 `} 
    - CancelUponError: whether to cancel a task upon execution error. A map of errorCode and true of false, by default all erros are true, meaning the task will be cancelled upon error. 
    { "errorCode": true or false}, for example, {"InsufficientBalance": false}

2. Next, the actual logic should run and generate an event for its result, such as , [parachainStake.delegationIncreased](https://turing.subscan.io/event?address=&module=parachainstaking&event=delegationincreased&startDate=&endDate=&startBlock=&endBlock=&timeType=date&version=294). This event should resemble the ones triggered by manual signing and should not be related to task events. The event can indicate success or failure, marking the execution result of the task. The execution event index should always be equal to the event index of the above TaskTrigger + 1. Currently, events have different names like XcmpTaskSucceeded and SuccessfullyAutoCompoundedDelegatorStake. We do not need to create special events like these; instead, we can rely on the existing events in the underlying pallets.
3. If the task is a recurring schedule, a TaskRescheduled event should subsequently be fired, indicating that a new task has been paid for and scheduled for the next run. If the user's wallet does not have sufficient fees to pay for the next execution, an error will occur at this step, but it will not stop the execution mentioned above.

These changes offer several benefits:

1. The firing of a task execution is clearly indicated through TaskTriggered, serving as a notification that the triggering function is functioning properly.
2. The success or failure of the execution is independent of the triggering notification mentioned above.
3. The execution logic is separated from the rescheduling process. For a fixed schedule, `TaskRescheduled` will not fire, while for recurring tasks, `TaskRescheduled` will fire upon successful events and may or may not fire upon errors. For example, if the AutoCompoundDelegationStake fails due to insufficient balance to pay for the next execution fee, the `TaskRescheduled` event should still be fired.

Implementing these changes will enhance the clarity and separation of task-related events, providing a more structured and meaningful representation of task execution and rescheduling.

imstar15:
Fix the issue of auto compound task:

1. When there’s not enough fee to pay for this execution, fail it with InsufficientFee error but do not stop the recurring tasks. At the next execution, we check it again to see if there’s enough fee for execution.

2. The task will stop when there’s no staking delegation within the user’s wallet. That means the user has unstaked all from collators.

Changes in pallet-parachain-staking:
https://github.com/OAK-Foundation/moonbeam/pulls


pallets/automation-time/src/tests.rs
```diff
@@ -2214,36 +2214,60 @@ fn auto_compound_delegated_stake_reschedules_and_reruns() {
...
+
+		let found_event = events().into_iter().find(|e| match e {
+			RuntimeEvent::AutomationTime(
+				crate::Event::SuccesfullyAutoCompoundedDelegatorStake { .. },
+			) => true,
+			_ => false,
+		});
+
+		if found_event.is_some() {
+			panic!("AutoCompound success eventAutoCompound success event should not be emitted.");
```
chrisli30, line 2269: This message looks wrong in grammar. Which is the success event name?

pallets/automation-time/src/tests.rs
```diff
@@ -2214,36 +2214,60 @@ fn auto_compound_delegated_stake_reschedules_and_reruns() {
 fn auto_compound_delegated_stake_without_minimum_balance() {
```
chrisli30, line 2214: Since we updated the logic, please write down the expected behavior for this test in comments.

pallets/automation-time/src/tests.rs
```diff
@@ -2252,7 +2276,7 @@ fn auto_compound_delegated_stake_does_not_reschedule_on_failure() {
 	new_test_ext(START_BLOCK_TIME).execute_with(|| {
```
chrisli30, line 2276: Doesn’t this test need update too? `auto_compound_delegated_stake_does_not_reschedule_on_failure`

chrisli30: We need a few more tests, for example, 
the next execution will not be scheduled when the wallet has left all delegations.

pallets/automation-time/src/lib.rs
```diff
@@ -1096,31 +1100,52 @@ pub mod pallet {
...
+								<T as Config>::WeightInfo::run_auto_compound_delegated_stake_task(),
+								None,
+							)
+						},
+						Err(e) => {
+							Self::deposit_event(Event::AutoCompoundDelegatorStakeFailed {
+								task_id,
+								error_message: Into::<&str>::into(e).as_bytes().to_vec(),
+								error: e.into(),
+							});
```
chrisli30, line 1140: **What message will emit** when you decide not to schedule the next exeuction when there’s no more delegations?

chrisli30 (changes requested):

pallets/automation-time/src/lib.rs
```diff
@@ -264,6 +266,8 @@ pub mod pallet {
 		BadVersion,
 		UnsupportedFeePayment,
 		CannotReanchor,
+		// Delegation Do Not Exist
+		DelegationDNE,
```
chrisli30, line 270: Never use three-letter acronym like `DNE`. Nobody knows what that means.

Please rename it to,
`DelegationNonFound`

pallets/automation-time/src/tests.rs
```diff
@@ -2252,7 +2276,7 @@ fn auto_compound_delegated_stake_does_not_reschedule_on_failure() {
...
 		get_funds(AccountId32::new(ALICE));
 		let before_balance = Balances::free_balance(AccountId32::new(ALICE));
-		let account_minimum = before_balance * 2;
+		let account_minimum = before_balance / 2;
```
chrisli30, line 2279: Can you explain this `/ 2` **in Rust code comments**?

pallets/automation-time/src/tests.rs
chrisli30: No.2
**auto_compound_delegated_stake_enough_balance_no_delegation**
Condition: wallet has no delegation with the specificed collator
Expected result: 
- current execution will not run
- Event will emit: AutoCompoundDelegatorFail, message: "wallet has no delegation with the specificed collator"
- next execution will not be scheduled

No.3
**auto_compound_delegated_stake_not_enough_balance_has_delegation**
condition: wallet balance < minimum balance
Expected result: 
- current execution will not run
- Event will emit: ExecutionSkipped, message: "Balance less than minimum"
- next execution will be scheduled

There are 4 scenarios
1. Wallet has enough balance & Delegation
**2. Wallet has enough balance & without Delegation
3. Wallet has not enough balance & Delegation**
4. Wallet has not enough balance & without Delegation

pallets/automation-time/src/tests.rs
imstar15: I have add test cases for 4 scenarios

pallets/automation-time/src/tests.rs

pallets/automation-time/src/tests.rs
```diff
@@ -2234,13 +2271,19 @@ fn auto_compound_delegated_stake_without_minimum_balance() {
...
 		AutomationTime::trigger_tasks(Weight::from_ref_time(120_000));
 
-		events()
+		let emitted_events = events();
+
+		// Expected result:
+		// 1. The current execution will be skipped, triggering the emission of an AutoCompoundDelegatorStakeSkipped event, message: Balance less than minimum
+		emitted_events.clone().into_iter()
+			.find(|e| matches!(e, RuntimeEvent::AutomationTime(crate::Event::AutoCompoundDelegatorStakeSkipped { message, .. }) if *message == "Balance less than minimum".as_bytes().to_vec())).expect("AutoCompoundDelegatorStakeSkipped event should have been emitted");
+
```
chrisli30, line 2280: Is it easy for the tests.rs file to reference the error messages in lib.rs?

For example, here you can use `*message == ERROR_MESSAGE_BALANCE_LOW` to refer to the message definition in lib.rs.

pallets/automation-time/src/tests.rs
chrisli30: Great, the test cases look clearer now.

chrisli30 (changes requested):
Comments for the new test cases are clear, 👍  however, there are a few more comments and questions.

Besides, we need to update the Documentation regarding this API on Docs website, https://docs.oak.tech/docs/yield-boost-auto-restaking/. @imstar15 please work on the documentation after this PR is approved and merged.

1. The skip logic needs to be added to the page
2. Error code and messages need to be listed on the page.

pallets/automation-time/src/lib.rs
```diff
@@ -1096,30 +1104,57 @@ pub mod pallet {
 			task: &TaskOf<T>,
 		) -> (Weight, Option<DispatchError>) {
 			// TODO: Handle edge case where user has enough funds to run task but not reschedule
+			if !T::DelegatorActions::is_delegation_exist(&delegator, &collator) {
+				let e = sp_runtime::DispatchErrorWithPostInfo::from(Error::<T>::DelegationNonFound);
+				Self::deposit_event(Event::AutoCompoundDelegatorStakeFailed {
+					task_id,
+					error_message: Into::<&str>::into(e).as_bytes().to_vec(),
```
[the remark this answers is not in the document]
chrisli30: Discussed with Charlse. In this line of code, error_message doesn’t contain more info than the `error` so we need to pass in variables to make it more meaningful. Let’s not bother the length for now. We can prune them later.

@v9n, there are two ways to go with this `error_message`,
First, we can add full information here, e.g. "delegator:<delegator_wallet_address>. collator: <collator_wallet_address>", however, the two wallet information can be inferred and don’t really needed in this case, because delegator is the wallet who created this task and collator is specified during task creation.

Adding the full error_message will help live debugging, but as you said will increase output and maybe blockchain data size. What do you recommend, should we log the full message or leave it blank? Could leaving it blank really save block space (for example, maybe the size of the event is decided by its largest parameter)?

pallets/automation-time/src/lib.rs
```diff
@@ -1145,7 +1149,13 @@ pub mod pallet {
 						},
 					}
 				},
-				None => (<T as Config>::WeightInfo::run_auto_compound_delegated_stake_task(), None),
+				None => {
+					Self::deposit_event(Event::AutoCompoundDelegatorStakeSkipped {
+						task_id,
+						message: "Balance less than minimum".into(),
```
chrisli30, line 1155: Let’s define a constant for these string messages.

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
chrisli30: @imstar15, as Vinh said, let’s add the variables to error_message in the below format, and update the tests accordingly.
"delegator:<delegator_wallet_address>, collator: <collator_wallet_address>"

pallets/automation-time/src/tests.rs
chrisli30: Yes, the `account_minium` variable name is self-explanatory. This is resolved.

pallets/automation-time/src/lib.rs
```diff
@@ -64,21 +64,32 @@ use frame_support::{
...
+		CheckedConversion, CheckedSub, Convert, Dispatchable, SaturatedConversion, Saturating,
+	},
 	ArithmeticError, DispatchError, Perbill,
 };
 use sp_std::{boxed::Box, vec, vec::Vec};
 pub use weights::WeightInfo;
 use xcm::{latest::prelude::*, VersionedMultiLocation};
 
+#[macro_export]
+macro_rules! ERROR_MESSAGE_DELEGATION_FORMAT {
```
imstar15, line 86: In Rust, the format! macro requires the formatting string to be a string literal. We cannot use constants or static variables directly as the formatting string. To address this, we can use a macro to extract the formatting string.

chrisli30: I don’t think we need an extra `hex` library to encode the wallets. Can you write down the actual message of the `error_message` here? Is the wallet address still human readable after hex encoding?

On second thought, if we were to add more details, we should always print out delegator and collator for all AutoCompoundDelegatorStake events in a way that is easier to parse.

For example, 
AutoCompoundDelegatorStakeSucceeded
- taskId
- amount
- collator

AutoCompoundDelegatorStakeSkipped
- taskId
- amount
- collator
- message: `Unable to pay fees for the next execution`
- value: <the expected amount to pay for the next execution>
> The balance reason needs an explanation, as in what was the balance of the wallet, so we add that information by adding an extra field to this event. The challenge is to find a common method to add extra information to a message, or an error. Let me ask round to find out the best practice to extend the Error definition. @v9n any recommendations? We need to standardize the format of Success and Error events.

AutoCompoundDelegatorStakeFailed
- taskId
- amount
- collator
- error: DelegationNotFound
> Since there’s already a collator information included, we don’t need any extra fields for this `AutoCompoundDelegatorStakeFailed` error

chrisli30 (changes requested):
The macro_export approach looks a little clumsy to me, so we need to find a better solution for Events and Error.

pallets/automation-time/src/lib.rs
imstar15: After undergoing `hex` processing, it will become an account ID string.
For example: 0x3830ef2fdb02d711afe91c830737375e32ac9904886266897f361b5a2ec13a39

chrisli30: Yes, wallet addresses need to be printed out, but not hex or accountId. @imstar15

pallets/automation-time/src/lib.rs
```diff
@@ -325,31 +330,23 @@ pub mod pallet {
...
+			who: AccountOf<T>,
 			task_id: TaskId<T>,
-			error_message: Vec<u8>,
-			error: DispatchErrorWithPostInfo,
-		},
-		AutoCompoundDelegatorStakeSkipped {
-			task_id: TaskId<T>,
-			message: Vec<u8>,
+			error: DispatchErrorWithDataMap,
 		},
```
chrisli30, line 337: Since AutoCompoundDelegatorStakeSucceeded and AutoCompoundDelegatorStakeSkipped are gone, could this error event be replaced by TaskExecutionFailed as well? 

We don’t need to define `DispatchErrorWithDataMap` either, because delegation information will be contained in the `TaskTriggered` event.

chrisli30: `TaskTriggered` is not defined yet.

pallets/automation-time/src/types.rs
```diff
@@ -213,6 +214,7 @@ pub struct Task<AccountId, Balance> {
 	pub provided_id: Vec<u8>,
 	pub schedule: Schedule,
 	pub action: Action<AccountId, Balance>,
+	pub cancel_upon_errors: Vec<String>,
 }
```
chrisli30, line 218: The name of `cancel_upon_errors` was not well thought. Let’s call it `abortErrors` (abort_errors) to be more intuitive.

Any task with a Recurring schedule should include this parameter.

The detailed explanation is in #393

pallets/automation-time/src/tests.rs
```diff
@@ -327,9 +329,19 @@ fn schedule_transfer_with_dynamic_dispatch() {
...
 		assert_eq!(
 			my_events,
 			[
+				RuntimeEvent::AutomationTime(crate::Event::TaskTriggered {
+					who: account_id.clone(),
+					task_id: task_id.clone(),
+					condition,
+					encoded_call: call.encode(),
+					abort_errors: vec![],
+				}),
```
chrisli30, line 344: The below two events are not necessary in our test. Remember this, for each operation we only need to test one event, which is usually the last event. For example in this test is `Event::Transfer`. This way our automation tests don’t need to be too coupled with the logic of other pallets, and testing `Event::Transfer` is sufficient to make sure the encoded_call happen.

```
RuntimeEvent::System(frame_system::Event::NewAccount {
					account: recipient.clone()
				}),
				RuntimeEvent::Balances(pallet_balances::pallet::Event::Endowed {
					account: recipient.clone(),
					free_balance: 127,
				}),
```

pallets/automation-time/src/tests.rs
```diff
@@ -1793,15 +1818,6 @@ mod run_dynamic_dispatch_action {
...
 			);
-
-			assert_eq!(
-				events(),
-				[RuntimeEvent::AutomationTime(crate::Event::DynamicDispatchResult {
-					who: account_id,
-					task_id: task_id.clone(),
-					result: Err(DispatchError::BadOrigin),
-				}),]
-			);
```
chrisli30, line 1804: Where’s the replacement of this error verification?

pallets/automation-time/src/tests.rs
```diff
@@ -1818,15 +1834,6 @@ mod run_dynamic_dispatch_action {
...
 			);
-
-			assert_eq!(
-				events(),
-				[RuntimeEvent::AutomationTime(crate::Event::DynamicDispatchResult {
-					who: account_id,
-					task_id: task_id.clone(),
-					result: Err(DispatchError::from(frame_system::Error::<Test>::CallFiltered)),
-				}),]
-			);
```
chrisli30, line 1829: Where’s the replacement of this error verification? What is this case testing without it?

chrisli30: The same applies to the below tests.

pallets/automation-time/src/tests.rs
```diff
@@ -2125,34 +2156,63 @@ fn trigger_tasks_limits_missed_slots() {
...
 				RuntimeEvent::AutomationTime(crate::Event::Notify { message: message_one.clone() }),
+				RuntimeEvent::AutomationTime(crate::Event::TaskExecuted {
+					who: owner.clone(),
+					task_id: task_id1.clone(),
+				}),
 				RuntimeEvent::AutomationTime(crate::Event::TaskCompleted {
-					who: AccountId32::new(ALICE),
+					who: owner.clone(),
 					task_id: task_id1.clone(),
 				}),
```
chrisli30, line 2201: `Event::TaskCompleted` is not supposed to fire here. It fires only after all executions of a task is done.

pallets/automation-time/src/tests.rs
```diff
@@ -2387,17 +2470,33 @@ fn missed_tasks_removes_completed_tasks() {
...
+					task_id: task_id01.clone(),
+					condition: condition.clone(),
+					encoded_call: vec![],
+					abort_errors: vec![],
+				}),
 				RuntimeEvent::AutomationTime(crate::Event::Notify { message: message_one }),
+				RuntimeEvent::AutomationTime(crate::Event::TaskExecuted {
+					who: owner.clone(),
+					task_id: task_id01.clone(),
+				}),
```
chrisli30, line 2492: If `TaskMissed` fires, the task is missed. There shouldn’t be `TaskTriggered` and `TaskExecuted`.

If it is the 2nd execution that is missed, add comments here to mark the start of the second task execution and explain the logic.

pallets/automation-time/src/tests.rs
```diff
@@ -2462,30 +2565,39 @@ fn trigger_tasks_completes_some_xcmp_tasks() {
...
 			[
-				RuntimeEvent::AutomationTime(crate::Event::XcmpTaskSucceeded {
-					destination,
+				RuntimeEvent::AutomationTime(crate::Event::TaskTriggered {
+					who: owner.clone(),
 					task_id: task_id.clone(),
+					condition,
+					encoded_call,
+					abort_errors: vec![],
 				}),
```
chrisli30, line 2595: Is it possible to verify the success event of the encoded_call here?

pallets/automation-time/src/tests.rs
```diff
@@ -2494,192 +2606,403 @@ fn trigger_tasks_completes_some_xcmp_tasks() {
...
-			delegation_event,
-			RuntimeEvent::AutomationTime(crate::Event::SuccesfullyAutoCompoundedDelegatorStake {
-				task_id,
-				amount: before_balance
-					.checked_sub(account_minimum.saturating_add(fee))
-					.expect("Event should not exist if value is neg"),
-			})
-		);
+			.find(|e| matches!(e, RuntimeEvent::AutomationTime(crate::Event::TaskExecuted { .. })))
+			.expect("TaskExecuted event should have been emitted");
```
chrisli30, line 2641: The format of the `events` for verification is different from other tests. Please update the format to be consistent with others.

pallets/automation-time/src/tests.rs
```diff
@@ -2494,192 +2606,403 @@ fn trigger_tasks_completes_some_xcmp_tasks() {
...
 			.into_iter()
-			.find(|e| match e {
-				RuntimeEvent::AutomationTime(
-					crate::Event::SuccesfullyAutoCompoundedDelegatorStake { .. },
-				) => true,
-				_ => false,
-			})
-			.expect("AutoCompound success event should have been emitted");
+			.find(|e| matches!(e, RuntimeEvent::AutomationTime(crate::Event::TaskTriggered { .. })))
+			.expect("AutoCompoundDelegatorStakeSucceeded event should have been emitted again!");
```
chrisli30, line 2720: Please make sure all the expect messages updated. Here it shouldn’t be `AutoCompoundDelegatorStakeSucceeded` any more.

pallets/automation-time/src/tests.rs
```diff
@@ -2776,32 +3133,62 @@ fn on_init_runs_tasks() {
...
+					condition: condition.clone(),
+					encoded_call: vec![],
+					abort_errors: vec![],
+				}),
 				RuntimeEvent::AutomationTime(crate::Event::Notify { message: message_one.clone() }),
+				RuntimeEvent::AutomationTime(crate::Event::TaskExecuted {
+					who: owner.clone(),
+					task_id: task_id1.clone(),
+				}),
 				RuntimeEvent::AutomationTime(crate::Event::TaskCompleted {
```
chrisli30, line 3176: `Event::TaskCompleted` shouldn’t fire here. It fires only at the very end.

chrisli30: The same applies to the below tasks.

chrisli30 (changes requested):

pallets/automation-time/src/tests.rs
imstar15: Yes, It indeed triggers when the task is completed.

There are 3 tasks in this test, and each task is executed only once. So, after each task is executed, the Event::TaskCompleted will be triggered.

pallets/automation-time/src/tests.rs
imstar15: No, this is an XCMP task, and its encoded call will not be executed on turing.

pallets/automation-time/src/tests.rs
imstar15: Yes, It indeed triggers when the task is completed.

There are 2 tasks in this test, and each task is executed only once. So, after each task is executed, the Event::TaskCompleted will be triggered.

pallets/automation-time/src/tests.rs
imstar15: Fixed

pallets/automation-time/src/tests.rs
chrisli30: okay, please add comments here to indicated that the first task is completed and the subsequent events are for the rest of tasks.

pallets/automation-time/src/tests.rs
chrisli30: okay, please add comments here to indicated that the first task is completed and the subsequent events are for the rest of tasks.

pallets/automation-time/src/tests.rs
imstar15: Fixed

pallets/automation-time/src/tests.rs
chrisli30: Low priority - this doesn’t need to block this PR. I created a https://github.com/OAK-Foundation/OAK-blockchain/issues/395 to address it later.

pallets/automation-time/src/tests.rs
```diff
@@ -510,17 +523,11 @@ fn will_emit_task_completed_event_when_task_failed() {
 		AutomationTime::trigger_tasks(Weight::from_ref_time(900_000_000));
 		let my_events = events();
 
-		// The DynamicDispatchResult event with error should be emitted when task failed.
```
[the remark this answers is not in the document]
imstar15: InsufficentBalance will not be thrown as an event during task execution.

Example:
https://turing.subscan.io/extrinsic/113696-2

it will be thrown as TaskExecutionFailed after the task fails.

pallets/automation-time/src/tests.rs
```diff
@@ -2364,6 +2446,7 @@ fn missed_tasks_removes_completed_tasks() {
 			vec![40],
```
chrisli30, line 2446: I created a #396 to track this test case not accurate issue in order to fix it later. It shouldn’t block this PR.

pallets/automation-time/src/tests.rs
imstar15: Fixed

pallets/automation-time/src/tests.rs
imstar15: Fixed

pallets/automation-time/src/tests.rs
```diff
@@ -523,11 +553,31 @@ fn will_emit_task_completed_event_when_task_failed() {
...
+					task_id: task_id.clone(),
+					condition,
+					encoded_call: call.encode(),
+					abort_errors: vec![],
+				}),
+				RuntimeEvent::AutomationTime(crate::Event::TaskExecutionFailed {
+					who: account_id.clone(),
+					task_id: task_id.clone(),
+					error: <pallet_balances::Error<Test>>::InsufficientBalance.into(),
+				}),
```
chrisli30, line 574: Nice, tests written in this format is much clearer 👍

pallets/automation-time/src/tests.rs
```diff
@@ -327,32 +365,35 @@ fn schedule_transfer_with_dynamic_dispatch() {
...
 					task_id: task_id.clone(),
-					result: Ok(()),
 				}),
 				RuntimeEvent::AutomationTime(crate::Event::TaskCompleted {
 					who: account_id,
 					task_id,
 				}),
 			]
-		);
+		));
```
chrisli30, line 396: @v9n, this list is the expected event firing behavior for successful task executions. The PR is ready for review. Could you double check it, especially the dev code in automation-time/src/lib.rs?

pallets/automation-time/src/lib.rs
```diff
@@ -1483,7 +1533,7 @@ pub mod pallet {
 			match task.schedule {
 				Schedule::Fixed { .. } =>
 					Self::decrement_task_and_remove_if_complete(task_id, task),
-				Schedule::Recurring { .. } => Self::reschedule_or_remove_task(task, error),
+				Schedule::Recurring { .. } => Self::reschedule_or_remove_task(task_id, task, error),
```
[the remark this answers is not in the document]
imstar15: Good call!
Fixed.

pallets/automation-time/src/lib.rs
```diff
@@ -360,6 +346,23 @@ pub mod pallet {
 			who: AccountOf<T>,
 			task_id: TaskIdV2,
 		},
+		TaskTriggered {
+			who: AccountOf<T>,
+			task_id: TaskIdV2,
+			condition: BTreeMap<String, String>,
+			encoded_call: Vec<u8>,
+			/// List of errors causing task abortion.
+			abort_errors: Vec<String>,
```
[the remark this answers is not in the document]
chrisli30: Good call,

TaskTriggered doesn’t need this array cuz it’s too noisy.

Neither does TaskNotRescheduled need it, because it already contains the actual error that caused the NotRescheduled event. We can always look up the task creation event to verify that the causing error is in the abort_errors array.

TaskRescheduleFailed doesn’t need this because that error is most likely to be caused by block full and not relate to abort_errors.

pallets/automation-time/src/lib.rs
```diff
@@ -316,23 +313,12 @@ pub mod pallet {
...
 			task_id: TaskIdV2,
 			execution_time: UnixTime,
 		},
-		/// The result of the DynamicDispatch action.
-		DynamicDispatchResult {
-			who: AccountOf<T>,
-			task_id: TaskIdV2,
-			result: DispatchResult,
-		},
 		/// The call for the DynamicDispatch action can no longer be decoded.
```
chrisli30, line 322: We don’t need this CallCannotBeDecoded event.

When CallCannotBeDecoded error happens, the below two Events need to capture it
1. TaskExecutionFailed
2. TaskNotRescheduled.

pallets/automation-time/src/lib.rs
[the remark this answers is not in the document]
imstar15: Fixed

pallets/automation-time/src/benchmarking.rs
```diff
@@ -324,9 +327,6 @@ benchmarks! {
 		let call: <T as Config>::Call = frame_system::Call::remark { remark: vec![] }.into();
 		let encoded_call = call.encode();
 	}: { AutomationTime::<T>::run_dynamic_dispatch_action(caller.clone(), encoded_call, task_id.clone()) }
-	verify {
```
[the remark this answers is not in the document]
imstar15: Because `run_dynamic_dispatch_action` does not throw events, it only returns error. Therefore, I checked the error it returns. Since error variable cannot be used in the `verify` block, I did not include these checks in the `verify` block

Refers to:
https://github.com/paritytech/substrate/blob/fbddfbd76c60c6fda0024e8a44e82ad776033e4b/frame/state-trie-migration/src/lib.rs#L964

.

pallets/automation-time/src/lib.rs
imstar15: Fixed

pallets/automation-time/src/mock.rs
```diff
@@ -556,9 +639,10 @@ pub fn add_task_to_missed_queue(
 	task_id: TaskIdV2,
 	scheduled_times: Vec<u64>,
 	action: ActionOf<Test>,
+	cancel_errors: Vec<String>,
```
chrisli30, line 642: Should this `cancel_errors` in this test case be updated?

pallets/automation-time/src/benchmarking.rs
```diff
@@ -316,26 +319,26 @@ benchmarks! {
...
 	}
 
 	run_dynamic_dispatch_action_fail_decode {
 		let caller: T::AccountId = account("caller", 0, SEED);
 		let task_id = vec![49, 45, 48, 45, 52];
 
 		let bad_encoded_call: Vec<u8> = vec![1];
-	}: { AutomationTime::<T>::run_dynamic_dispatch_action(caller.clone(), bad_encoded_call, task_id.clone()) }
-	verify {
-		assert_last_event::<T>(Event::CallCannotBeDecoded{ who: caller, task_id: task_id.clone() }.into())
```
chrisli30, line 338: Maybe I’m wrong, but Is this `assert_last_event` util function recently added by @v9n? Should we keep or remote the function?

chrisli30 (approved):
Looks good enough to me.

pallets/automation-time/src/mock.rs
imstar15: This does not need to be removed. Because we need to record the abort_errors of each task in order to know how to terminate the task.

pallets/automation-time/src/benchmarking.rs
imstar15: Yes, we don't need it now.

If left there, a warning will be generated. Perhaps this warning can be removed through some annotations.

pallets/automation-time/src/mock.rs
imstar15: Yes, Fixed!

pallets/automation-time/src/benchmarking.rs
imstar15: I revert assert_last_event and add  allow(dead_code) annotation on it

PR #384, opened 2023-07-17
Auto generated task id and remove provided_id

v9n:
This PR change how task id of `automationTime` pallet is generated. We removed `provided_id` concept, and generate a custom `task_id`. 

- Refactor our task id generation to guarantee all task ids are unique globally moving forward.
- Improve consistency on our extrinsic/rpc: there is only `task id`.

### Before

Original, when schedule a task user need to pass a `provided_id`, and the task id will be the hash of `(wallet_address, provided_id)`. The task id when emit can be re-present in hex format such as:

```
0x68c6fc86f787fe825e67a1299657154313e4662a3aef9deb7eaec10632154da7
```

### After

The task id is auto generated and is an Vec<u8> using this formula:

```
{block number}-{extrinsic index in the block}-{event count at the time the extrinsic is executed}
```

In practice it looks like this

```
2080853-12-93
```

The old task id that we already generated and computed will be convert into Vec<u8> using its hex re-presentation. Example, if a task id before has its hex is `0x68c6fc86f787fe825e67a1299657154313e4662a3aef9deb7eaec10632154da7` then the task id will be the literal string(including the 0x prefix) `0x68c6fc86f787fe825e67a1299657154313e4662a3aef9deb7eaec10632154da7` .

So when interacting with our RPC/Extrinsic, client can pass the string `2080853-12-93` or `0x68c6fc86f787fe825e67a1299657154313e4662a3aef9deb7eaec10632154da7`  as the task id.

Example of the new task id format:

Screenshot 2023-07-25 at 5 59 30 PM

### scheduleDynamicDispatchTask

Before:

Screenshot 2023-07-25 at 11 52 14 AM

After:

Screenshot 2023-07-25 at 5 48 03 PM

### scheduleXcmpTask

Before:

Screenshot 2023-07-25 at 11 52 48 AM

After:

Screenshot 2023-07-25 at 5 50 11 PM

### scheduleXcmpTaskThroughProxy

Before:

Screenshot 2023-07-25 at 12 07 21 PM

After:

Screenshot 2023-07-25 at 5 50 53 PM

### Removed RPC

Below RPCs are removed. The client can listen to the change event TaskScheduled to find the task id of a task. There i sno need to look up task id using provided id anymore.

- automationTime.generateTaskId

# Proposal doc

https://docs.google.com/document/d/1fyHsuXp98usuSN3ltJ_klox9zT0yY2hqX7i0C7hkvnc/edit#heading=h.180hrep5fk9q

### DIFF
--- a/pallets/automation-time/src/migrations/update_task_idv2.rs
+++ b/pallets/automation-time/src/migrations/update_task_idv2.rs
@@ -21,14 +21,8 @@ use xcm::{latest::prelude::*, VersionedMultiLocation};
 
 use crate::migrations::utils::{
 	deprecate::{generate_old_task_id, old_taskid_to_idv2},
-	OldAccountTaskId, OldMissedTaskV2Of, OldScheduledTasksOf,
+	OldAccountTaskId, OldMissedTaskV2Of, OldScheduledTasksOf, TEST_TASKID1, TEST_TASKID2,
 };
-
-// These are H256/BlakeTwo256 hex generate from our old task id generation from hashing
-// These cons are used for our unit test
-// (Account, ProvidedID)
-const TEST_TASKID1: &str = "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04";
-const TEST_TASKID2: &str = "0xf76acf0b1d8ef503450a4d5c1a451f1921a906e8711f551c2945e09fb44de5ff";
 
 // This is old TaskQueueV2 with old Task
 #[frame_support::storage_alias]
@@ -256,22 +250,8 @@ mod test {
 			let task_owner2 = AccountId32::new(BOB);
 
 			// These are H256/BlakeTwo256 hex generate from our old task id generation from hashing
-			// (Account, ProvidedID)
-			// 0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04
-			let task_id1: Vec<u8> = vec![
-				48, 120, 100, 49, 50, 54, 51, 56, 52, 50, 101, 51, 52, 97, 100, 101, 98, 48, 48,
-				98, 101, 49, 49, 52, 54, 98, 51, 48, 98, 99, 54, 53, 50, 55, 57, 53, 49, 102, 48,
-				97, 48, 102, 53, 100, 53, 97, 51, 102, 55, 97, 55, 53, 56, 53, 51, 55, 55, 51, 53,
-				98, 56, 98, 99, 98, 48, 52,
-			];
-			// 0xf76acf0b1d8ef503450a4d5c1a451f1921a906e8711f551c2945e09fb44de5ff
-			// (account_id2, 2) ->
-			let task_id2: Vec<u8> = vec![
-				48, 120, 102, 55, 54, 97, 99, 102, 48, 98, 49, 100, 56, 101, 102, 53, 48, 51, 52,
-				53, 48, 97, 52, 100, 53, 99, 49, 97, 52, 53, 49, 102, 49, 57, 50, 49, 97, 57, 48,
-				54, 101, 56, 55, 49, 49, 102, 53, 53, 49, 99, 50, 57, 52, 53, 101, 48, 57, 102, 98,
-				52, 52, 100, 101, 53, 102, 102,
-			];
+			let task_id1: Vec<u8> = TEST_TASKID1.as_bytes().to_vec();
+			let task_id2: Vec<u8> = TEST_TASKID2.as_bytes().to_vec();
 
 			super::ScheduledTasksV3::<Test>::insert(
 				3600,
@@ -338,22 +318,16 @@ mod test {
 					// (task owner, vec![33]) hash -> "0x7191f3d83bcbeb221f7b00f501e9a8da3ba3c2d15a672eb694ee7e09dbaddd1e"
 					(
 						task_owner2.clone(),
-						vec![
-							48, 120, 55, 49, 57, 49, 102, 51, 100, 56, 51, 98, 99, 98, 101, 98, 50,
-							50, 49, 102, 55, 98, 48, 48, 102, 53, 48, 49, 101, 57, 97, 56, 100, 97,
-							51, 98, 97, 51, 99, 50, 100, 49, 53, 97, 54, 55, 50, 101, 98, 54, 57,
-							52, 101, 101, 55, 101, 48, 57, 100, 98, 97, 100, 100, 100, 49, 101
-						]
+						"0x7191f3d83bcbeb221f7b00f501e9a8da3ba3c2d15a672eb694ee7e09dbaddd1e"
+							.as_bytes()
+							.to_vec(),
 					),
-					// (task owner1, vec![32]) hash -> "0x7191f3d83bcbeb221f7b00f501e9a8da3ba3c2d15a672eb694ee7e09dbaddd1e"
+					// (task owner1, vec![32]) hash -> "0xe94040ca5d09f0e1023aecf5abc3afac0b9e66bc3b1209183b3a009f4c073c2b"
 					(
 						task_owner1.clone(),
-						vec![
-							48, 120, 101, 57, 52, 48, 52, 48, 99, 97, 53, 100, 48, 57, 102, 48,
-							101, 49, 48, 50, 51, 97, 101, 99, 102, 53, 97, 98, 99, 51, 97, 102, 97,
-							99, 48, 98, 57, 101, 54, 54, 98, 99, 51, 98, 49, 50, 48, 57, 49, 56,
-							51, 98, 51, 97, 48, 48, 57, 102, 52, 99, 48, 55, 51, 99, 50, 98
-						]
+						"0xe94040ca5d09f0e1023aecf5abc3afac0b9e66bc3b1209183b3a009f4c073c2b"
+							.as_bytes()
+							.to_vec(),
 					),
 				],
 				"migration failed to convert old ScheduledTasksV3 to new TaskID format"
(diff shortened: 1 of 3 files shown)


pallets/automation-time/src/benchmarking.rs
```diff
@@ -354,9 +360,9 @@ benchmarks! {
 		for i in 0..v {
 			let provided_id: Vec<u8> = vec![i.saturated_into::<u8>()];
 			let task = TaskOf::<T>::create_event_task::<T>(caller.clone(), provided_id.clone(), vec![time.into()], vec![4, 5, 6]).unwrap();
-			let task_id = AutomationTime::<T>::schedule_task(&task, provided_id).unwrap();
-			let missed_task = MissedTaskV2Of::<T>::new(caller.clone(), task_id, time.into());
-			<AccountTasks<T>>::insert(caller.clone(), task_id, task);
+			let task_id = AutomationTime::<T>::schedule_task(&task).unwrap();
+			let missed_task = MissedTaskV2Of::<T>::new(caller.clone(), task_id.clone(), time.into());
+			<AccountTasks<T>>::insert(caller.clone(), task_id.clone(), task);
```
chrisli30, line 365: I notice there are several addition of the `.clone()`. Is the `task_id.clone()` to create a deep clone to prevent the original variable from getting destroyed outside the insert() call?

pallets/automation-time/src/mock.rs
```diff
@@ -598,6 +589,27 @@ pub fn last_event() -> RuntimeEvent {
 	events().pop().unwrap()
 }
 
+pub fn last_task_id() -> TaskIdV2 {
+	get_task_ids_from_events()
+		.last()
+		.expect("task id is missing from event. maybe job schedule failed")
```
chrisli30, line 595: This will fail only if the return of `get_task_ids_from_events()`, right? If so, the message should be `Unable to find a task_id from the existing TaskScheduled events`.

pallets/automation-time/src/mock.rs
```diff
@@ -598,6 +589,27 @@ pub fn last_event() -> RuntimeEvent {
 	events().pop().unwrap()
 }
 
+pub fn last_task_id() -> TaskIdV2 {
```
chrisli30, line 592: This utility function looks good, however, could you add some comments to explain its purpose? Like, why using this approach is better than the existing code?

runtime/oak/src/lib.rs
```diff
@@ -1246,7 +1242,10 @@ impl_runtime_apis! {
 					AutomationTime::generate_task_id(account_id.clone(), provided_id)
 				})
 				.filter(|task_id| {
-					AutomationTime::get_account_task(account_id.clone(), task_id).is_some()
+					// TODO(Vinh): migration
+					let mut t = vec![];
```
chrisli30, line 1246: I see several `t` variables in this PR. Let’s never use a variable name that has no meaning. `tasks` or `tasksTemp` could make sense.

chrisli30 (commented):
Looks good in general. I left a few comments.

pallets/automation-time/src/lib.rs
```diff
@@ -1475,6 +1518,30 @@ pub mod pallet {
...
+					Some(i) => i,
+					None => 0,
+				};
+
+			let tx_id = match <frame_system::Pallet<T>>::extrinsic_index() {
+				Some(i) => i,
+				None => 0,
+			};
+
+			let evt_index = <frame_system::Pallet<T>>::event_count();
```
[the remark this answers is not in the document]
imstar15: The change in event_count depends on the fact that the TaskScheduled event of the previous task will be emitted.

 I am somewhat concerned about whether this generation method of task id is secure and reliable enough.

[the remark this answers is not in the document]
imstar15: After giving it some thought, I believe this approach should be fine. When a task is successfully registered, the TaskScheduled event will definitely be emitted, resulting in an increment of the event count. Following this logic, it should be correct, and thus, taskId conflicts should not occur.

The only scenario where a problem may arise is if the TaskScheduled event is not emitted, which would be a separate issue to address. However, based on the current logic, it should work as expected.

pallets/automation-time/src/lib.rs
```diff
@@ -1344,19 +1355,34 @@ pub mod pallet {
 			.map_err(|_| Error::<T>::TimeSlotFull)
 		}
 
+		// TODO: move to a utils upon this MVP works
+		// We cannot use Rust std inside substrate runtime so i need to re-invent this
+		// Convert a decimal number to its string re-presentation where each character is a digit
+		// of the original number
```
chrisli30, line 1361: hmm it’s strange that the code base lacks of basic utils like this. Let’s find a home for this util function in this PR. @imstar15 where do we currently store utility functions?

pallets/automation-time/src/lib.rs
```diff
@@ -1344,19 +1355,34 @@ pub mod pallet {
...
+				s.push((48 + digit) as u8);
+				if b <= 0 {
+					break
+				}
+			}
+			s.reverse();
+			s
+		}
+
 		/// TODO ENG-538: Refactor validate_and_schedule_task function
```
chrisli30, line 1378: ENG-538: Refactor validate_and_schedule_task function is deleted on Jira. Could you help remove thes lines of function comment?

pallets/automation-time/src/types.rs
```diff
@@ -446,10 +435,10 @@ mod tests {
 					vec![0],
 				)
 				.unwrap();
-				let task_id = TaskId::<Test>::default();
+				let task_id = vec![48, 45, 48, 45, 48];
```
chrisli30, line 438: maybe use the same thing TaskId default?

pallets/automation-time/src/benchmarking.rs
chrisli30: Ah I see. Okay, hmm yeah changing underlining signagure is too big of a change. I got it. Let’s keep them as-is.

pallets/automation-time/src/tests.rs
```diff
@@ -43,6 +43,10 @@ pub const START_BLOCK_TIME: u64 = 33198768000 * 1_000;
 pub const SCHEDULED_TIME: u64 = START_BLOCK_TIME / 1_000 + 7200;
 const LAST_BLOCK_TIME: u64 = START_BLOCK_TIME / 1_000;
 
+// This is 1-0-3: {1: block idx}-{0: first extrinsic in block}-{3: the event index}
```
chrisli30, line 46: Comments look good 👍

pallets/automation-time/src/migrations/update_task_idv2.rs
```diff
@@ -0,0 +1,441 @@
...
+	#[test]
+	fn on_runtime_upgrade() {
+		new_test_ext(0).execute_with(|| {
+			let account_id1 = AccountId32::new(ALICE);
+			let account_id2 = AccountId32::new(BOB);
+
+			// These are H256/BlakeTwo256 hex generate from our old task id generation from hashing
+			// (Account, ProvidedID)
+			// 0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04
+			let task_id1: Vec<u8> = vec![
```
imstar15, line 220: Please reuse the string literal "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04" as follows:

```
const TASK_ID1: &str = "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04";
const TASK_ID2: &str = "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04";

let task_id1: Vec<u8> = TASK_ID1.as_bytes().to_vec();
let task_id2: Vec<u8> = TASK_ID2.as_bytes().to_vec();
```

By reusing the same string literal for `TASK_ID1` and `TASK_ID2`, we can ensure consistency and avoid duplication of the same data.

imstar15, line 220: Please reuse the string literal "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04" as follows:

on the top of the file:
```
const TASK_ID1: &str = "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04";
const TASK_ID2: &str = "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04";
```

in the function:
```
let task_id1: Vec<u8> = TASK_ID1.as_bytes().to_vec();
let task_id2: Vec<u8> = TASK_ID2.as_bytes().to_vec();
```

By reusing the same string literal for `TASK_ID1` and `TASK_ID2`, we can ensure consistency and avoid duplication of the same data.

imstar15 (approved):
Looks good to me!

pallets/automation-time/src/migrations/update_task_idv2.rs
```diff
@@ -0,0 +1,428 @@
...
+	fn on_runtime_upgrade_for_missed_queue_v2() {
+		new_test_ext(0).execute_with(|| {
+			let account_id1 = AccountId32::new(ALICE);
+			let account_id2 = AccountId32::new(BOB);
+
+			// These are H256/BlakeTwo256 hex generate from our old task id generation from hashing
+			// (Account, ProvidedID)
+			// (account_id1, 1) ->
+			// 0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04
+			let task_id1: Vec<u8> = vec![
```
imstar15, line 374: Please search "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04" and replace with string literal.

pallets/automation-time/src/migrations/update_xcmp_task.rs
```diff
@@ -189,16 +148,26 @@ mod test {
...
-			super::AccountTasks::<Test>::insert(account_id.clone(), task_id, task);
+			let old_task_id =
+				generate_old_task_id::<Test>(account_id.clone(), task.provided_id.clone());
+			super::AccountTasks::<Test>::insert(account_id.clone(), old_task_id.clone(), task);
 
 			UpdateXcmpTask::<Test>::on_runtime_upgrade();
 
 			assert_eq!(crate::AccountTasks::<Test>::iter().count(), 1);
+
+			// 0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04
```
imstar15, line 159: Please search "0xd1263842e34adeb00be1146b30bc6527951f0a0f5d5a3f7a758537735b8bcb04" and replace with string literal.

imstar15 (changes requested):

imstar15 (approved):
Looks good to me!

PR #390, opened 2023-07-25
Minor readme update and clean up outdated zombienet config

chrisli30:
This PR contains the below updates
1. The `zombienets/turing/xcmp.toml` file relies on a fork of substrate-template code repo, which is not maintained any more. In addition, since we have xcmp configs with real partners, we no longer need a self-maintained fake parachain. Therefore, I’m deleting the xcmp.toml file.
2. Renaming the `zombienets/turing/single-chain.toml` filename to `zombienets/turing/turing-simple.toml` because it’s specifically for Turing Dev. The reason it’s called simple is because I plan to add another turing config file to test staking. The current `turing-simple.toml` does not come with AURA consensus for staking.
3. Small README.md correction.


imstar15 (approved):
Looks good to me!

PR #391, opened 2023-07-26
Update the format of turing srtool output in the release template

chrisli30:
The current value is hard to read so I updated the format.

**Before:**
Version     : turing-294 (oak-1.tx17.au1)

**After:**
Spec version     : 294
Impl version     : 1
Tx version       : 17

`authoring_version` is omitted because we will probably never change it.


imstar15 (approved):
Looks good to me!

PR #397, opened 2023-08-01
add ParachainStaking into our allowed schedule

v9n:
Currently only Balances/System are allowed to be schedule through dynamic dispatch. This is an attempt to allow us call schedule for ParachainStaking extrinsic

Screenshot 2023-08-01 at 3 20 21 PM


chrisli30 (approved):
okay, looks good. We will test parachainStaking automation manually on Turing Staging.

PR #398, opened 2023-08-02
fix xcmp migration fail to decoded error

v9n:
Currently we iterate, migrate and insert(replace) migrated task in the same iteration. There is a chance we will iterate over the item we just added, and it will not be able to decode use the old Rust struct.

Therefore we do migrations in 2 steps. Copy out all the item, transform them, cleanup the storage and re-insert later.


chrisli30 (approved):
LGTM

PR #400, opened 2023-08-03
Fix xcmp return weight

v9n:
the old xcmp migration task return weight from `migration_upgrade_xcmp_task`. this weight definition is being removed from our benchmark.rs in lastcode, so upon re-generate benchmark, we will lost it. 

Plus the way substrate mention is count the amount of ops we do to storage. when we loop over the task, we read/insert them back. there is one extra time where we access the whole Storage itself.


chrisli30 (approved):
Yes, this makes sense to me. 🆗

PR #403, opened 2023-08-05
fix calculate task id in benchmark verify block

v9n:
@chrisli30 @imstar15  sorry for the last minute PR. I discovered this when attempting to run the benchmark.

depend on where we run, in unit test or in benchmark, the amount of initial event will be different, instead of hard code a number as before which make the verify cannot pass both of benchmark/unitest, we dynamically get the event count from initial events in the block to calculate task id.

this way the `verify` works for both of benchmark/unittest

Upon fixing both of unitest are pass and benchmark is run succesfully locally

```
Min Squares Analysis
========
-- Extrinsic Time --

Model:
Time ~=       16
              µs

Reads = 1
Writes = 1
Recorded proof Size = 8942

Created file: "./pallets/automation-time/src/raw-weights.rs"
```


chrisli30 (approved):
oh yeah this totally makes sense 👌

PR #405, opened 2023-08-05
Update README.md to add instructions of running the latest zombienet with polkadot 0.9.43

chrisli30:
Since relay chain’s node startup command changed, the old zombienet source code stopped working. I have tested and added instructions of using [zombienet v1.3.63](https://github.com/paritytech/zombienet/releases/tag/v1.3.63)

**The command that works** on my Macbook Pro is, `./zombienet-macos spawn zombienets/turing/mangata.toml`

Specifically, the polkadot change that broke it is,
CleanShot 2023-08-04 at 19 01 15

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@
 # Parachain registration files
 genesis-state
 genesis-wasm
+
+# zombienet binary for MacOS to run scripts in the ./zombienets folder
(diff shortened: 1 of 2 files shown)


imstar15 (approved):
Looks good to me!

PR #407, opened 2023-08-06
Update zombienet script to fix port numbers of relay chain nodes

chrisli30:
This is a minor change to use fixed relay chain port numbers instead of randomly assigned when using zombienet. Using fixed ports will allow us to have consistent port numbers upon zombienet startup, instead of looking for a random number from the command logs.

This PR needs a rebase once #405 is approved and merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@
 # Parachain registration files
 genesis-state
 genesis-wasm
+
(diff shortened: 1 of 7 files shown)


imstar15 (approved):
Looks good to me!

PR #410, opened 2023-08-07
Remove notify and native transfer

v9n:
Complely remove reference of Notify/NativeTransfer enum and action in our code.

For those tasks, we attempt to convert them to dynamic dispatch task. For `Notify` we used `System::remark_with_event` it ins't 100% compatible with our old notify because they emit the Hash of the string, where as our old Notify emit the string.


chrisli30:
I’ll leave the review to @imstar15.

imstar15 (approved):

PR #412, opened 2023-08-08
use Vector of old task id hash for new task id

v9n:
instead of converting the hash to its hex format and use the ascii string of the hex as Vec<u8> for new task id, we use the original [u8, 32] of the Hash and convert it to a Vec<u8>

This will keep our task id show up in the same way before and after update.


chrisli30 (approved):
Cool, looks simpler

PR #413, opened 2023-08-09
Remove XcmpTaskFailed, TransferFailed event in automation-time pallet

imstar15:


chrisli30 (approved):
Looks good. No need to include this in 2.0.0 since the release is already generated.

Issue #414, opened 2023-08-09
Clean up Github Actions


chrisli30:
Vinh, since you separated Action script recently, I’m assigning this item to you. There seems to be several Github Actions script deprecated. Could you check them out in the below screenshot to see if my requirements make sense?
![CleanShot 2023-08-08 at 20 14 48](https://github.com/OAK-Foundation/OAK-blockchain/assets/2616844/4382fe77-4f7d-4998-9b06-64bb302c4de0)

Regarding the renaming, I think "Publish Turing Docker Image" is good enough. In addition, it would be also great to include the docker image name `oaknetwork/turing` in its title or description.

chrisli30:
Okay, got it. Let’s keep the actions, and make the name consistent then, by renaming **"Relase Docker Image"** to **"Publish Turing Docker Image"**, because very soon there will be another, "Publish OAK Docker Image".

PR #415, opened 2023-08-09
cleanup action and refactor docker publish

v9n:
cleanup unused workflow actions.

Add a "stealth release" mode that push the image to docker hub, but won't push the `latest` pointer to avoid being picked up by our node operator that use docker auto update.

2.0.0 is pushed but `latest` has no changed.

Screenshot 2023-08-09 at 4 06 17 PM

Screenshot 2023-08-09 at 4 05 18 PM

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -3,7 +3,7 @@ name: Relase Docker Image
 on:
   workflow_dispatch:
     inputs:
-      relase_action_run_id:
+      release_action_run_id:
         required: true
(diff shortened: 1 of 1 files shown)


chrisli30 (approved):
Looks good to me. The draft mode is a good idea 👍 

One thing off of the top of my head, a common practice for us is to add "Fix #414" to either the title or the description of a PR if it’s to fix an issue. That way the issue will be automatically closed when the PR is merged. However, we don’t need that in this PR since it doesn’t fix the #414 alone.

PR #417, opened 2023-08-22
Remove changelog.yml from .github/workflow

chrisli30:
@imstar15 we don’t use this file, right? I don’t see any successful Actions run. If not, we can delete this deprecated file.


imstar15 (approved):
Yes, we can delete it.

PR #419, opened 2023-08-29
cleanup migration code from v2.0.0

v9n:
part of post migration to cleanup migration that already ran and applied.


chrisli30 (approved):
Great, Vinh!

PR #421, opened 2023-09-11
Clippy & cleanup

siddharthteli12:
Code refactoring no functional changes.

### DIFF
--- a/pallets/automation-price/src/lib.rs
+++ b/pallets/automation-price/src/lib.rs
@@ -304,7 +304,7 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(_: T::BlockNumber) -> Weight {
-			if Self::is_shutdown() == true {
+			if Self::is_shutdown() {
 				return T::DbWeight::get().reads(1u64)
 			}
 
@@ -344,8 +344,7 @@ pub mod pallet {
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 			let fee = <BalanceOf<T>>::saturated_from(1_000_000_000_000u64);
-			T::FeeHandler::can_pay_fee(&who, fee.clone())
-				.map_err(|_| Error::<T>::InsufficientBalance)?;
+			T::FeeHandler::can_pay_fee(&who, fee).map_err(|_| Error::<T>::InsufficientBalance)?;
 			Self::validate_and_schedule_task(
 				who.clone(),
 				provided_id,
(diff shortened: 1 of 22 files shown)


siddharthteli12:
This pr is ready for review. There's further scope for optimisation which will require some additional time. Can start working on that if this pr is accepted & maintainer feel its helpful.

imstar15:
Hi @siddharthteli12 

Some errors were encountered while compiling

https://github.com/OAK-Foundation/OAK-blockchain/actions/runs/6147646245/job/16699352425?pr=421

siddharthteli12:
@imstar15 Will resolve it.

siddharthteli12:
@imstar15 ,Not sure how I missed that error as I build the node. But nevertheless fixed that typo🙂.

imstar15:
May I ask, what Clippy command line are you using? Is it 'cargo clippy --fix'?

siddharthteli12:
Yes, Initially ran `cargo clippy --fix` then did some manual update.

chrisli30:
The unit tests have passedin the All checks. @v9n please take a look at the result of this tool `clippy` as well to see if the suggestions are correct.

imstar15:
Looks good to me!

siddharthteli12:
@v9n, Adding `clippy` to action & 2 pr idea makes sense.

Also, The changes to `BountyDepositPayoutDelay` was by `clippy --fix`.  I agree the latter is more readable there's some configuration to make `clippy` more lenient. Will update on that soon.

siddharthteli12:
@v9n , I did some research on clippy & I think adding auto `clippy --fix` option in actions will lead to problems. Running `cargo clippy --fix` will auto fix only a subset of lint for others it will keep on throwing errors causing GitHub action to break. 

My suggestion would be to add `cargo clippy` in GitHub action (If we really want to add it) & set a standard to run `cargo clippy --fix` before  `pushing` which might also require some manual fixes as `fix` can't act on all error/warning. 

Apart from this, `clippy fix` tent to change weight file for eg - `reads(6 as u64))` to `reads(6_u64))` which is not a big issue but as `clippy` doesn't natively support substrate it will enforce such changes.

siddharthteli12:
Reverted unit multiplier like in `BountyDepositPayoutDelay: BlockNumber = 1 * DAYS` for better readability. 

@v9n , I would suggest to wait for a while for code refactor before adding clippy to GitHub action. I see considerable clippy logs which should be sorted first. Personally, I believe clippy should not be enforced with GitHub actions as it is optioned

siddharthteli12:
@v9n , Sure it does make sense. 
As you also said above,  We can start by cleaning current lint logs. After that clippy in action will make sure all future commit are lint free & follow a standard.

PR #428, opened 2023-09-16
Update weight template

siddharthteli12:
With ref. to #421 & #427. 

-  Weight  [template update](https://github.com/OAK-Foundation/OAK-blockchain/blob/master/.maintain/frame-weight-template.hbs) which was also causing unnecessary logs in `clippy`. 
- Also, why `weights.rs` exist but benchmarking file is missing for [automation_price](https://github.com/OAK-Foundation/OAK-blockchain/tree/master/pallets/automation-price/src)?


imstar15:
> * Also, why `weights.rs` exist but benchmarking file is missing for [automation_price](https://github.com/OAK-Foundation/OAK-blockchain/tree/master/pallets/automation-price/src)?

Yes
I believe the benchmarking file is indeed missing, and we should add it.

siddharthteli12:
For now, Just updated the weights for automation price as per the latest template. Later we can add benchmarks.
Overall, we have some consistency now. So, I believe this pr is ready for review.

siddharthteli12:
Weight template is already updated in master. Just updated the weights in this pr. It's ready for review.

siddharthteli12:
Will close this pr.

PR #429, opened 2023-09-18
Remove refund instructions

imstar15:
The reason of these removal is that each XCM instruction costs 1M weight, so removing the last two for refunding actually will cost users less than having a refund process.

Besides, Rocstar doesn’t allow high XCM weight per block, so reducing weight of each XCM messages will help our messages get unblocked.


chrisli30 (approved):
Yeah @imstar15 please explain the PR when you submit it ...

@v9n, the reason of these removal is that each XCM instruction costs 1M weight, so removing the last two for refunding actually will **cost users less** than having a refund process.

Besides, Rocstar doesn’t allow high XCM weight per block, so reducing weight of each XCM messages will help our messages get unblocked.

PR #430, opened 2023-09-18
Increase max_message_size in Shibuya zombienet config to avoid polkadotXcm.SendFailure error

imstar15:


chrisli30 (approved):
okay, good good

Issue #434, opened 2023-09-25
Price Triggering


v9n:
A Place holder to keep track of progress and TODO to be fully consider the feature is completed

- [x] Add benchmark and update weight for automation-price pallet
- [x] Cleanup TODO
- [x] Handle delete asset extrinsic
- [x] Improve error handling
- [ ] Optimize price trigger performance: only scan/check asset that has price actually move in the same directon with the triggering function.
- [x] General code cleanup
- [ ] Simplify interface API(There will be a goole doc next to discuss those) https://docs.google.com/document/d/1JIKL0BrKAZTKWzeNdoMrtb9OJ6elDmePCpDBaUlVejY/edit

## Add relevant event similar to Automation Time
  - [x] TaskCancelled
  - [x] TaskCompleted
  - [x] TaskExecuted
  - [x] TaskExecutionFailed

Sub PR:

https://github.com/OAK-Foundation/OAK-blockchain/pull/420

PR #435, opened 2023-09-27
add benchmark for automation-price

v9n:
Starting adding properly benchmark to prepare for weight and fee handler.

Part of #434

### DIFF
--- a/pallets/automation-price/src/lib.rs
+++ b/pallets/automation-price/src/lib.rs
@@ -162,14 +162,6 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxWeightPercentage: Get<Perbill>;
 
-		/// The maximum number of authorized wallet address to update price for an asset
-		#[pallet::constant]
-		type MaxAuthorizedOracleWallet: Get<u32>;
-
-		/// The maximum element of prices vector in price update extrinsics
-		#[pallet::constant]
-		type MaxBatchPriceUpdate: Get<u32>;
-
 		#[pallet::constant]
 		type ExecutionWeightFee: Get<BalanceOf<Self>>;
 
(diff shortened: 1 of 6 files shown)


pallets/automation-price/src/lib.rs
```diff
@@ -314,12 +324,10 @@ pub mod pallet {
 
 	#[pallet::error]
 	pub enum Error<T> {
-		/// The provided_id cannot be empty
 		EmptyProvidedId,
```
chrisli30, line 327: This Error related to the providedId is the old clientId pre-2.0.0. Is it being used by automationPrice?

runtime/neumann/src/lib.rs
```diff
@@ -929,6 +929,11 @@ impl pallet_automation_price::Config for Runtime {
 	type UniversalLocation = UniversalLocation;
 	type SelfParaId = parachain_info::Pallet<Runtime>;
 	type EnsureProxy = AutomationEnsureProxy;
+
+	// how many oracle can be authorized to update price for a pair
+	type MaxAuthorizedOracleWallet = ConstU32<5>;
+	// How many item we can put into the same price update extrinsic
+	type MaxBatchPriceUpdate = ConstU32<100>;
```
chrisli30, line 936: These are good ideas. The variable names are self-explanatory. Since they are new, we will need to document them along with other new events or chain configs on docs.oak.tech. 

I suggest us to keep it simple for now since we will be using a single oracle for the next 3-6 months, because these two parameters will need a set of unit tests to ensure the edge cases or errors are handled correctly

chrisli30 (approved):
No major issues identified, and I love the code comments 👍

PR #436, opened 2023-09-28
add fee handling for automation price

v9n:
Now that we have weight, attempt to handle fee properly. 

We charged fee only once when schedule the job succesfully. Because this isn't recurring tasks, we only need to charge once.


pallets/automation-price/src/tests.rs
```diff
@@ -407,13 +444,14 @@ fn test_shift_tasks_movement_through_price_changes() {
 			Box::new(NATIVE_LOCATION.into()),
 			Box::new(AssetPayment {
 				asset_location: MultiLocation::new(0, Here).into(),
-				amount: 10000000000000
+				amount: 10
```
chrisli30, line 447: Is 10 too small for testing purpose? 10 * 10^10 units would make more sense?

chrisli30 (changes requested):
I re-compiled with this PR but found tasks are not being scheduled after Shibuya’s polkadotXcm.send. 

@v9n could you test the code with our demo to see what’s wrong? https://github.com/OAK-Foundation/automation-price-demo

chrisli30 (commented):
I have no more comments on this PR. @imstar15 could you also examine the XCM related code?

pallets/automation-price/src/tests.rs
chrisli30: @v9n

pallets/automation-price/src/lib.rs
```diff
@@ -756,12 +765,9 @@ pub mod pallet {
 			};
 
 			Self::validate_and_schedule_task(task)?;
-			// TODO withdraw fee
-			//T::FeeHandler::withdraw_fee(&who, fee).map_err(|_| Error::<T>::InsufficientBalance)?;
```
chrisli30, line 760: Without this line, is the fee actually withdrawn?

chrisli30 (approved):
LGTM. I left a few comments.

PR #437, opened 2023-09-29
add event when task is triggered and executed and canceled

v9n:
Add 3 events similar to automation time.

There is one exception is that automation time has TaskCompleted event when all of its execution is done. In our procing based model, the task is consider completed once we had emitted `TaskExecuted`.

For sake of consistency, we may consider to emit the extra TaskCompleted event later.


chrisli30 (approved):
The code looks good to me.

Yes, we should emit TaskCompleted for consistency, and that event is a clear indication that the task has been removed from the task queue.

PR #439, opened 2023-09-30
Fix #438 Handle asset deletion

### LINKED ISSUE #438: automationPrice.deleteAsset doesn’t seem to delete an asset properly
**Steps to Reproduce**
1. Submit an extrinsic to delete an asset
Encoded call hash: 0xdfb4ba0568a5b12c4abe449f9d7f155bbfea51ed96b923103f98a7926e3cce91

CleanShot 2023-09-29 at 21 23 58

2. The AssetDelete event shows up, but only with one asset in its details.
CleanShot 2023-09-29 at 21 24 33

3. The original asset is still in chain storage
CleanShot 2023-09-29 at 21 24 53

4. Although the delete extrinsic succeeded, re-create the same asset will get an "automationPrice.AssetAlreadyInitialized" error during extrinsic submission.
initAsset Encoded call hash: 0xc8011c73686962757961206172746873776170145752535452105553445412040c720beb3f580f0143f9cb18ae694cddb767161060850025a57a4f72a71bf475

v9n:
Handle asset removal. There are 2 scenarios:

- there is no tasks for this asset. we will delete asset pair immediately
- if there is pending task, we will put this deletion into schedule for removal and delete later on.

### DIFF
--- a/pallets/automation-price/src/lib.rs
+++ b/pallets/automation-price/src/lib.rs
@@ -462,6 +462,7 @@ pub mod pallet {
 			decimal: u8,
 			asset_owners: Vec<AccountOf<T>>,
 		) -> DispatchResult {
+			// TODO: use sudo and remove this feature flag
 			// TODO: needs fees if opened up to non-sudo
 			// When enable dev-queue, we skip this check
 			#[cfg(not(feature = "dev-queue"))]
@@ -573,6 +574,7 @@ pub mod pallet {
 			asset1: AssetName,
 			asset2: AssetName,
 		) -> DispatchResult {
+			// TODO: use sudo and remove this feature flag
 			// When enable dev queue, we want to skip this root check so local development can
 			// happen easier
 			#[cfg(not(feature = "dev-queue"))]


chrisli30:
Fix #438 (@v9n, please add this fix clause to your PR title or description for linking and auto-resolving after merge)

Thanks Vinh for the quick fix! My thoughts on this topic after checking out this PR.

First and foremost, when an asset is deleted, it indicates that the pool associated with it is either deprecated or no longer supported by our chain. In either case, any tasks that rely on this asset should be canceled as well. Additionally, since there won't be any price updates after the asset's deletion, any pending tasks that have not yet run will never execute. Therefore, it is essential to cancel all tasks dependent on this asset's price to free up valuable chain space.

Thinking about the scenarios, I found there are many edge cases around the concept of `AssetScheduledToDelete`. For simplicity I suggest us to get rid of the scheduler concept and sweep relevant pending tasks right away. We just need to be cautions not to sweep at a busy block. **Alternatively**, we could offer a more flexible solution by allowing manual sweeps through batches of forceCancelTask(ownerId, taskId) calls, without embedding the sweep logic directly into the blockchain. 

With the manual sweep solution, the deletion logic becomes much simpler. However, we must also implement forceCancelTask(ownerId, taskId) for automationPrice.

chrisli30:
> Yes, here is some complex logic around pending tasks thus I didn't implement it at first place and want to give more though later on.
> 
> I agree we should sweep it right away instead of scheduling to be deleted. When we delete an asset, there are either:
> 
> * we make a mistake at first place
> * Or we stopped supporting it, that mean we will need to stop that price oracle as well.
> 
> In either case, clearly having the deletion done asap is desired. If we keep it around and schedule to be deleted later on, the price oracle needs to be around too.
> 
> So I propose:
> 
> 1. If there is no pending task, immediately remove asset and price etc
> 2. If there are pending task, return error say "Having pending tasks etc..."
> 
> It's the caller duty to cleanup the task at first place before deleting it. We will introduce 2 type of functions
> 
> * forceCancelTaskByIds: sudo(government) can cancel a list of task by its id.
> * forceCancelTaskByAssets: sudo(government) can cancel a list of task by assets
> 
> Note: originally reason we have the `forceCancelTask(owner_id, task_id)` is because the owner is parted of the task id. In our new format, the task id alone is enough to find the task. So when cancel by, knowing task_id is enough.
> 
> With this model, the caller need to sweep all the tasks manually before deletion and they know what they are doing, such as run it at a less busy time.
> 
> When the forceCancelTask run, we do this:
> 
> * delete as much as possible. If there is still pending task, we will emit an event `TaskCancelProgress { done: 100, remain: 200 }` for example. The caller will know if there is any remaining to deleteagain.
> 
> Because technically asset removeal isn't something we done on day to day basic, therefore I opted for above model where the caller need to explicitly handle this task sweep.
> 
> When all tasks are clean, the caller call remove it.

I agree with your design above, but let me add more context to it. 
1. Yes, the forceCancelTask needs only one parameter, which is `taskId`.
2. As you mentioned, the `force` function is made for `sudo` permission, so there’s only one caller of this function, which is the council. Asset creator has no permission to delete anybody else’s tasks.
3. Given that in mind, as long as we as the council is carefully specifying no more than 100 taskIds at once, one extrinsic `forCancelTasks(taskIds)` should suffice, eliminating the need of introducing of a new event, TaskCancelProgress. 
4. ***Important**: The asset functions **should not depend on** task execution logic, or vice versa. The asset deletion should work before or after task cancellations. Please test before and after scenarios in tests. 
5. In my opinion this is a low hanging fruit, so **my estimate is one day to complete this asset deletion PR and add a forCancelTasks(taskIds) PR to achieve what are needed with minimum effort**. Let’s ship the product to market with simple solutions first, and document these sophiscated ideas for later development.

chrisli30:
In summary, let’s do the below things quickly. Let me know if they make sense to you?
- Clean up and test AssetDeleted, and remove AssetScheduledToDelete. The asset deletion does not need to interfere with any tasks. 
- Add a `forCancelTasks(taskIds)` in a separate PR, for the council to cancel tasks related to the delete asset. The taskIds will be retrieved from insights dashboard manually. 
- There’s no need for `TaskCancelProgress` or related logic at the moment.

chrisli30 (approved):
LGTM

PR #448, opened 2023-10-08
Remove get_time_automation_fees RPC function

imstar15:
Because the query_fee_details function returns both schedule_fee and execution_fee, while get_time_automation_fees only returns schedule_fee.

we should keep only query_fee_details and remove get_time_automation_fees.


chrisli30 (approved):
Yes, we discussed this and the get_time_automation_fees was an early experiment which is deprecated now.

PR #451, opened 2023-10-10
Sweep expired tasks, fixed #442

### LINKED ISSUE #442: Expired tasks stay in tasks queue and not get cleaned
The expiredAt value works well, as tasks are not executed once the block time passes it. However, expired tasks stay in `tasks` result and are not get cleaned up. It’s not a high-pri but we definitely need to have means to clean them.

v9n:
When block has idle time, we will run task sweep to delete expired tasks. We will run as long as we still have weight left leverage `on_idle` block https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.Hooks.html#method.on_idle

I had a few test cases to test this extensively. When removing tasks, we also update our stat such as total task count and task count of user that has task being removed.

### DIFF
--- a/pallets/automation-price/src/lib.rs
+++ b/pallets/automation-price/src/lib.rs
@@ -371,7 +371,6 @@ pub mod pallet {
 		TaskRemoveFailure,
 		/// Task Not Found When canceling
 		TaskNotFound,
-		TaskDoesNotExist,
 		/// Error when setting task expired less than the current block time
 		InvalidTaskExpiredAt,
 		/// Error when failed to update task expiration storage


pallets/automation-price/src/lib.rs
```diff
@@ -1232,8 +1253,145 @@ pub mod pallet {
...
 		}
 
-		fn remove_task(task: &Task<T>) {
-			//AccountTasks::<T>::remove(task.owner_id.clone(), task.task_id.clone());
+		// Handle task removal. There are a few places task need to be remove:
+		//  Tasks storage
+		//  TaskQueue if the task is alreayd queue
+		//  TaskStats: decrease task count
+		//  AccountStats: decrease task count
+		//  SortedTasksIndex
```
chrisli30, line 1261: Let’s add comments.

SortedTasksIndex - sorted by price
Sorted TaskIds by expiration?

PR #452, opened 2023-10-12
Fix #445 included bound on gt condition

### LINKED ISSUE #445: When a price is set at 100, schedule a task with condition > 100 will trigger immediately
## Actual
When the current price = 100, a task scheduled with condition > 100 will trigger immediately. 

Conversely, if **the same task** is created when price = 80, it is not be triggered when price is moved to 100 (which is the correct behavior).

CleanShot 2023-10-06 at 13 54 26

The condition is "gt", 100, shown in the encoded call of task creation:
```
0xc8051c73686962757961206172746873776170145752535452105553445458b02965000000000000000000000000086774046400000000000000000000000000000003010100411f03000003010100411febe1aee309491d00000000000000000089051200000c720beb3f580f0143f9cb18ae694cddb767161060850025a57a4f72a71bf475010040007c13030000000000000000000000000000000000000000000000000000000000a17e7ba271dc2cc12ba5ecf6d178bf818a6d76eb0000c16ff286230000000000000000000000000000000000000000000000000091037ff36ab5000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000005446cff2194e84f79513acf6c8980d6e60747253000000000000000000000000000000000000000000000000018abef7846071c700000000000000000000000000000000000000000000000000000000000000020000000000000000000000007d5d845fd0f763cefc24a1cb1675669c3da62615000000000000000000000000e17d2c5c7761092f31c9eca49db426d5f2699bf000076b14ea3f010e8c0300076b3c552e020e8c13000c720beb3f580f0143f9cb18ae694cddb767161060850025a57a4f72a71bf475
```

## Repro
This problem can be reproduced by https://github.com/OAK-Foundation/automation-price-demo

v9n:
fix #445 

When finding the task to execute, we use a range, I mistakenly using `Included` so the upper bound is included in the range, correct to Excluded and add a test for it.

### DIFF
--- a/pallets/automation-price/src/tests.rs
+++ b/pallets/automation-price/src/tests.rs
@@ -823,8 +823,6 @@ fn test_sweep_expired_task_partially() {
 					.map_or(0, |v| v.len()),
 			);
 		}
-=======
->>>>>>> 058648e (Fix > condition to exclude the upper bound)
 	})
 }
 


chrisli30 (approved):
Yes, if we call the condition `gt` it can’t include the boundary. We can use these for now and add conditions such as greater than or equal to later.

PR #453, opened 2023-10-12
Fix #443 Add extra data to reason about task trigger

### LINKED ISSUE #443: TaskTriggered event doesn’t containt a condition attribute indicating why a task is triggered
In automationPrice, the TaskTriggered event doesn’t contain enough data to indicate the trigger condition. Please refer to https://github.com/OAK-Foundation/OAK-blockchain/issues/387 to implement it similar to the TaskTriggered in AutomationTime

Event Name: TaskTriggered Attributes:
	•	taskId:
	•	who:
	•	condition: { type:time, value: 1689903366 }

v9n:
Fix #443 to add a new `condition` into TaskTrigger. This contains 3 pieices:

- task trigger func
- task trigger targe price
- asset price

### DIFF
--- a/pallets/automation-price/src/lib.rs
+++ b/pallets/automation-price/src/lib.rs
@@ -396,9 +396,9 @@ pub mod pallet {
 	#[derive(Debug, Encode, Eq, PartialEq, Decode, TypeInfo, Clone)]
 	pub enum TaskCondition {
 		TargetPriceMatched {
-            // record the state of the asset at the time the task is triggered
-            // when debugging we can use this to reason about why did the task is trigger
-	        chain: ChainName,
+			// record the state of the asset at the time the task is triggered
+			// when debugging we can use this to reason about why did the task is trigger
+			chain: ChainName,
 			exchange: Exchange,
 			asset_pair: AssetPair,
 			price: u128,
@@ -1123,9 +1123,9 @@ pub mod pallet {
 				if price_matched_target_condtion {
 					return (
 						Some(TaskCondition::TargetPriceMatched {
-                            chain: task.chain.clone(),
-                            exchange: task.exchange.clone(),
-                            asset_pair: task.asset_pair.clone(),
+							chain: task.chain.clone(),
+							exchange: task.exchange.clone(),
+							asset_pair: task.asset_pair.clone(),
 							price: this_task_asset_price.amount,
 						}),
 						consumed_weight,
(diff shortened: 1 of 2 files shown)


pallets/automation-price/src/lib.rs
```diff
@@ -399,6 +399,14 @@ pub mod pallet {
 	/// Many fields on this struct is optinal to support multiple error condition
 	#[derive(Debug, Encode, Eq, PartialEq, Decode, TypeInfo, Clone)]
 	pub enum TaskCondition {
+		TargetPriceMatched {
+			// The target price the task set to
+			target_price: u128,
+			trigger_function: Vec<u8>,
```
chrisli30, line 405: Okay, I think there might be confusion in this `condition` paramter, , possibly due to its ambiguity. The goal of the event is not to re-iterate the trigger condition itself, but to provide real-time value of the circumstances that triggered a task in the code. For example, the TaskTriggered event should look like

`TaskTriggered {who, taskId, condition: condition_struct }`
in which, the `condition_struct` mirrors the same structure as `{destination, exchange, asset1, asset2, amount}` with `amount` being the current value on chain, as a proof of trigger.

Keep in mind the definition of condition_struct is likely to change, so we should use the definition as much as possible, rather than deconstructing it, to faciliate future updates.

chrisli30 (changes requested):

pallets/automation-price/src/lib.rs
chrisli30: Okay, I think we can omit the task details from the event, since it’s trivial compared to the real-time on-chain price. The reason is because task detais do not change after creation, but price varies across blocks. We almost always need to leverage the Insights service to read the task details from the past.

chrisli30: In other words, the task details already have on-chain proof and can be retrieved from Insights, but the triggering price of this moment is not recorded yet.

chrisli30 (approved):
Okay, the changes look good. I’ll need to give it a try to see how accessible is the new definition of condition in client’s code, for example, maybe it’s as easy as `event.condition.toJson()`. 

Comparing its counter-party in automationTime’s code, I think these predefined structs provides a better framework for defining conditions. Let’s try this now, and update the code in automationTime later.
https://github.com/OAK-Foundation/OAK-blockchain/blob/23433654b2c512aed336a9ae6e4174230b74969c/pallets/automation-time/src/lib.rs#L892C2-L892C2

PR #454, opened 2023-10-12
Fix #440 Standardize event name

### LINKED ISSUE #440: Attribute definition discrepency in automationPrice.tasks v.s. system events
When calling extrinsics and querying system events, I find it’s cumbersome but to have two sets of names. For example, in events of automationPrice, `who` and `taskId` are used, but in data storage, it becomes `taskId, ownerId`.

TaskScheduled Event
{ who, taskId }

automationPrice.tasks
 {
      ownerId: 6757gffjjMc7E4sZJtkfvq8fmMzH2NPRrEL3f3tqpr2PzXYq
      taskId: 1549-0-1
      ...
}

I think we should define a standard for those names. 
1. The `who` is not a good name, because both `owner` and `scheduleAs` are "who"s. We should just call the creator of the task, "owner". 
2. The "taskId" seems good so far. 

A reminder that if we change the "who" to "owner" in events, we will need to update automationTime and data Insights service as well.

v9n:
Address #440 
At this moment, I only applied this change to automation-price first. Eventually  we will need to do the same for automation-time and update insight dashboard and our GraphQL document.

In this PR scope, I just want to refactor the new code (automation price) that we know won't break anything else.

### DIFF
--- a/pallets/automation-price/src/benchmarking.rs
+++ b/pallets/automation-price/src/benchmarking.rs
@@ -315,7 +315,7 @@ benchmarks! {
 	}: {
 		// remove a task at the end to simulate the worst case
 		AutomationPrice::<T>::remove_task(&task, Some(crate::Event::<T>::TaskSweep {
-			who: task.owner_id.clone(),
+			owner_id: task.owner_id.clone(),
 			task_id: task.task_id.clone(),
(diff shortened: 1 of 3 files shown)


chrisli30 (approved):
Yep, I agree. Let’s test out this format in automationPrice first. Code actually looks a little simpler with these changes.

PR #455, opened 2023-10-13
Fix #444 increase round when update price, refactor format

### LINKED ISSUE #444: Each entry in priceRegistry needs a unique id
From developer’s perspective, every data type in on-chain storage needs a unique id, otherwise there’s no way to know whether a Price has been updated. Please refer to the screenshot below, as you can see from the events the client couldn’t tell what server version the data entry is. 
priceRegistry-no-unique-id

v9n:
Fix #444 

We will increase the round by one when updating the price
Added `updated_at`
Rename `.amount` -> `.value`

### DIFF
--- a/pallets/automation-price/src/lib.rs
+++ b/pallets/automation-price/src/lib.rs
@@ -929,16 +929,9 @@ pub mod pallet {
 				if let Some(mut tasks) = Self::get_sorted_tasks_index(&key) {
 					let current_price = current_price_wrap.unwrap();
 
-					//Eg sell order, sell when price >
-					let range;
-					if trigger_func == TRIGGER_FUNC_GT.to_vec() {
-						range = (Excluded(&u128::MIN), Excluded(&current_price.value))
-					} else {
-						// Eg buy order, buy when price <
-						range = (Included(&current_price.value), Excluded(&u128::MAX))
-					};
-
-					for (&price, task_ids) in (tasks.clone()).range(range) {
+					for (&price, task_ids) in
+						(tasks.clone()).range(range_by_trigger_func(&trigger_func, &current_price))
+					{
 						// Remove because we map this into task queue
 						tasks.remove(&price);
 						let ref mut t = &mut (task_ids.clone());
--- a/pallets/automation-price/src/trigger.rs
+++ b/pallets/automation-price/src/trigger.rs
@@ -1,5 +1,13 @@
 use crate::{Config, PriceData, Task};
 
+use sp_std::{
+	collections::btree_map::Range,
+	ops::{
+		Bound,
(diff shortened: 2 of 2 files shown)


pallets/automation-price/src/lib.rs
```diff
@@ -1115,9 +1132,9 @@ pub mod pallet {
 				//
 				let price_matched_target_condtion =
 					if task.trigger_function == TRIGGER_FUNC_GT.to_vec() {
-						task.trigger_params[0] < this_task_asset_price.amount
+						task.trigger_params[0] < this_task_asset_price.value
 					} else {
-						task.trigger_params[0] > this_task_asset_price.amount
+						task.trigger_params[0] > this_task_asset_price.value
 					};
```
chrisli30, line 1138: I feel like for code re-use purpose there needs to be a wrapper function for the `is_price_condition_matched()`, since the price is compared earlier and here again for checking `PriceAlreadyMoved`. Moreover, when there are more than 2 enums in future, this `if else` condition needs to change.

pallets/automation-price/src/lib.rs
```diff
@@ -1115,9 +1132,9 @@ pub mod pallet {
 				//
 				let price_matched_target_condtion =
 					if task.trigger_function == TRIGGER_FUNC_GT.to_vec() {
-						task.trigger_params[0] < this_task_asset_price.amount
+						task.trigger_params[0] < this_task_asset_price.value
```
chrisli30, line 1135: so if the trigger_function is "GT", meaning the actual value is greater than the set target. 

The logic has no proble, but writing this line like 
```
if GT
    this_task_asset_price.value > task.trigger_params[0]
``` 
would be more intuitive ...

chrisli30 (approved):
The code looks good as-is.

pallets/automation-price/src/lib.rs
```diff
@@ -929,16 +929,9 @@ pub mod pallet {
...
-					if trigger_func == TRIGGER_FUNC_GT.to_vec() {
-						range = (Excluded(&u128::MIN), Excluded(&current_price.value))
-					} else {
-						// Eg buy order, buy when price <
-						range = (Included(&current_price.value), Excluded(&u128::MAX))
-					};
-
-					for (&price, task_ids) in (tasks.clone()).range(range) {
+					for (&price, task_ids) in
+						(tasks.clone()).range(range_by_trigger_func(&trigger_func, &current_price))
```
chrisli30, line 933: Is `range_by_trigger_func`more efficient than `for task in tasks { task.is_price_condition_match }`? I feel both functions the same purpose.

chrisli30 (approved):
The changes look good to me.

PR #460, opened 2023-10-27
Cancel task by schedule_as account

imstar15:


pallets/automation-time/src/lib.rs
```diff
@@ -589,6 +591,36 @@ pub mod pallet {
...
+
+			let task = AccountTasks::<T>::get(owner_id, task_id.clone())
+				.ok_or(Error::<T>::TaskDoesNotExist)?;
+
+			if !matches!(task.clone().action, Action::XCMP { schedule_as: Some(ref s), .. } if s == &who)
+			{
+				return Err(Error::<T>::TaskScheduleAsNotMatch.into())
+			}
+
+			Self::remove_task(task_id, task);
```
chrisli30, line 621: I mentioned that we need to TODO here but it actually should be in the remove_task wrapper(), so no action here.

The TODO is about refund reserved execution fees.

pallets/automation-time/src/tests.rs
```diff
@@ -508,6 +508,199 @@ fn will_not_emit_task_completed_event_when_task_canceled() {
...
+
+		// Assert that the task is still present in accountTasks.
+		assert_eq!(
+			matches!(task.clone().action, Action::XCMP { schedule_as: Some(ref s), .. } if s == &schedule_as),
+			true
+		);
+	})
+}
+
+// Cancel task with incorrect schedule_as and non-existent taskid will fail
```
chrisli30, line 622: The function comments are good, but there’s a catch with this one. The "schedule_as" is not "incorrect", right? Let’s update it to,

Calling `cancel_task_with_schedule_as` with a non-existent taskId will cause TaskDoesNotExist error.

pallets/automation-time/src/tests.rs
```diff
@@ -508,6 +508,199 @@ fn will_not_emit_task_completed_event_when_task_canceled() {
...
+				RuntimeOrigin::signed(schedule_as),
+				task_owner.clone(),
+				task_id.clone(),
+			),
+			Error::<Test>::TaskDoesNotExist
+		);
+	})
+}
+
+// Cancel task with incorrect schedule_as and incorrect owner will fail
```
chrisli30, line 645: Same as above. I don’t see the schedule_as incorrect from the code. Please update the comment to,

Calling cancel_task_with_schedule_as with an incorrect owner will cause `TaskDoesNotExist` error.

Also, the Error `TaskDoesNotExist` looks not accurate because the task with `taskId` does exist. It’s just that `owner` and `scheduleAs` don’t match. This is because we changed taskId format, that a taskId could uniquely identify a task. No change is required for now, since the dev code is logically correct.

chrisli30 (approved):
@v9n, could you also double check on this? After some e2e testing, we will need you to replicate the same cancel_with_scheduleAs logic to automationPrice pallet later

pallets/automation-time/src/tests.rs

pallets/automation-time/src/tests.rs

pallets/automation-time/src/lib.rs
imstar15: Fixed

pallets/automation-time/src/lib.rs
```diff
@@ -1296,6 +1296,9 @@ pub mod pallet {
 			}
 
 			AccountTasks::<T>::remove(task.owner_id.clone(), task_id.clone());
+
+			// Refund reserved execution fees
```
chrisli30, line 1300: No, @imstar15 you need to add a TODO tag to the comment. Otherwise the comment is explaining the code in the line below.

`// TODO: Add refund reserved execution fees here`

pallets/automation-time/src/tests.rs
```diff
@@ -619,7 +619,7 @@ fn cancel_task_with_incorrect_schedule_as_will_fail() {
 	})
 }
 
-// Cancel task with incorrect schedule_as and non-existent taskid will fail
+// Cancel task with schedule_as and non-existent taskid will fail
```
chrisli30, line 622: "will fail" is not specific enough. It’s better to declare the error here. Please use the full sentence in my previous comment.

chrisli30 (changes requested):

pallets/automation-time/src/tests.rs

pallets/automation-time/src/lib.rs
imstar15: Fixed

chrisli30 (approved):
All comments are addressed. Good to merge.

PR #461, opened 2023-11-02
Specify the currency for the execution fee

imstar15:
We have implemented the functionality to specify different currencies as the execution fee.

Based on whether the currency is native or foreign, we have employed different processing flows and XCM instructions.

### DIFF
--- a/pallets/automation-time/src/fees.rs
+++ b/pallets/automation-time/src/fees.rs
@@ -150,6 +150,8 @@ where
 		log::debug!(target: "FeeHandler", "FeeHandler::withdraw_fee, self.schedule_fee.asset_location: {:?}, self.schedule_fee.amount: {:?}",
 			self.schedule_fee.asset_location, self.schedule_fee.amount);
 		// Withdraw schedule fee
+		// When the expected deduction amount, schedule_fee_amount, is not equal to zero, execute the withdrawal process;
+		// otherwise, there’s no need to deduct.
 		if !self.schedule_fee.amount.is_zero() {
 			let currency_id = T::CurrencyIdConvert::convert(self.schedule_fee.asset_location)
 				.ok_or("InconvertibleMultilocation")?;
@@ -172,6 +174,8 @@ where
 					.ok_or("InconvertibleMultilocation")?;
 
 				let execution_fee_amount = execution_fee.amount;
+				// When the expected deduction amount, execution_fee_amount, is not equal to zero, execute the withdrawal process;
+				// otherwise, there’s no need to deduct.
 				if !execution_fee_amount.is_zero() {
 					T::XcmpTransactor::pay_xcm_fee(
 						currency_id,
--- a/pallets/automation-time/src/fees.rs
+++ b/pallets/automation-time/src/fees.rs
@@ -19,6 +19,7 @@
 use crate::{AccountOf, Action, ActionOf, Config, Error, MultiBalanceOf, Pallet};
 
 use frame_support::traits::Get;
+use frame_system::RawOrigin;
 use orml_traits::MultiCurrency;
 use pallet_xcmp_handler::{InstructionSequence, XcmpTransactor};
 use sp_runtime::{
@@ -301,6 +302,74 @@ mod tests {
 			);
 			assert_eq!(result.expect("success"), "called");
 			assert_eq!(has_callback_run, true);
+		})
+	}
+
+	#[test]
+	fn call_pay_checked_fees_for_with_normal_flow_and_foreign_schedule_fee_success() {
+		new_test_ext(0).execute_with(|| {
+			let destination = MultiLocation::new(1, X1(Parachain(PARA_ID)));
+			let alice = AccountId32::new(ALICE);
+			let mut has_callback_run = false;
+			let _ = Currencies::update_balance(
+				RawOrigin::Root.into(),
+				alice.clone(),
+				FOREIGN_CURRENCY_ID,
+				XmpFee::get() as i64,
+			);
+			fund_account(&alice, 900_000_000, 1, Some(0));
+
+			let action = Action::XCMP {
+				destination,
+				schedule_fee: destination.into(),
+				execution_fee: AssetPayment { asset_location: NATIVE_LOCATION.into(), amount: 10 },
+				encoded_call: vec![3, 4, 5],
+				encoded_call_weight: Weight::from_parts(100_000, 0),
+				overall_weight: Weight::from_parts(200_000, 0),
+				schedule_as: None,
+				instruction_sequence: InstructionSequence::PayThroughSovereignAccount,
+			};
+
+			let result = <Test as crate::Config>::FeeHandler::pay_checked_fees_for(
+				&alice,
+				&action,
+				1,
+				|| {
+					has_callback_run = true;
+					Ok("called")
+				},
+			);
+			assert_eq!(result.expect("success"), "called");
+			assert_eq!(has_callback_run, true);
+		})
+	}
+
+	#[test]
+	fn call_pay_checked_fees_for_with_normal_flow_and_foreign_schedule_fee_will_throw_insufficent_balance(
+	) {
+		new_test_ext(0).execute_with(|| {
+			let destination = MultiLocation::new(1, X1(Parachain(PARA_ID)));
+			let alice = AccountId32::new(ALICE);
+			fund_account(&alice, 900_000_000, 1, Some(0));
+
+			let action = Action::XCMP {
+				destination,
+				schedule_fee: destination.clone().into(),
+				execution_fee: AssetPayment { asset_location: NATIVE_LOCATION.into(), amount: 10 },
+				encoded_call: vec![3, 4, 5],
+				encoded_call_weight: Weight::from_parts(100_000, 0),
+				overall_weight: Weight::from_parts(200_000, 0),
+				schedule_as: None,
+				instruction_sequence: InstructionSequence::PayThroughSovereignAccount,
+			};
+
+			let result = <Test as crate::Config>::FeeHandler::pay_checked_fees_for(
+				&alice,
+				&action,
+				1,
+				|| Ok(()),
+			);
+			assert_err!(result, Error::<Test>::InsufficientBalance);
 		})
 	}
 
--- a/pallets/automation-time/src/fees.rs
+++ b/pallets/automation-time/src/fees.rs
@@ -130,6 +130,7 @@ where
 						self.schedule_fee.asset_location,
 						self.schedule_fee.amount,
 					)?;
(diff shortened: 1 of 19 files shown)


pallets/automation-time/src/fees.rs
```diff
@@ -39,11 +39,18 @@ pub trait HandleFees<T: Config> {
 	) -> Result<R, DispatchError>;
 }
 
+#[derive(Clone)]
+pub struct FeePayment<T: Config> {
+	pub asset_location: MultiLocation,
+	pub amount: MultiBalanceOf<T>,
+	pub is_local_deduction: bool,
```
chrisli30, line 46: Let’s rename this to `is_local`, and add the below comment in code. 

`// Determined by automation flow - normal or alternative`

pallets/automation-time/src/fees.rs
```diff
@@ -73,57 +80,89 @@ where
 {
 	/// Ensure the fee can be paid.
 	fn can_pay_fee(&self) -> Result<(), DispatchError> {
-		let fee = self.schedule_fee_amount.saturating_add(self.execution_fee_amount);
+		let mut additional_fee = MultiBalanceOf::<T>::zero();
+
```
chrisli30, line 84: Let’s add the purpose of each step to code comment.

pallets/automation-time/src/fees.rs
```diff
@@ -39,11 +39,18 @@ pub trait HandleFees<T: Config> {
...
+pub struct FeePayment<T: Config> {
+	pub asset_location: MultiLocation,
+	pub amount: MultiBalanceOf<T>,
+	pub is_local_deduction: bool,
+}
+
 pub struct FeeHandler<T: Config, TR> {
 	owner: T::AccountId,
 	pub schedule_fee_location: MultiLocation,
 	pub schedule_fee_amount: MultiBalanceOf<T>,
```
chrisli30, line 52: Let’s use the FeePayment type for `schedule_fee` too.

pallets/automation-time/src/fees.rs
imstar15: Fixed

pallets/automation-time/src/lib.rs
```diff
@@ -267,7 +275,10 @@ pub mod pallet {
 		/// The version of the `VersionedMultiLocation` value used is not able
 		/// to be interpreted.
 		BadVersion,
+		UnsupportedFeePayment,
```
chrisli30, line 278: All Error code needs detailed comment to explain its purpose. Please see the existing comments above.

pallets/automation-time/src/fees.rs
```diff
@@ -71,59 +78,110 @@ where
...
 		Ok(())
 	}
 
 	/// Withdraw the fee.
 	fn withdraw_fee(&self) -> Result<(), DispatchError> {
-		let fee = self.schedule_fee_amount.saturating_add(self.execution_fee_amount);
+		log::debug!(target: "FeeHandler", "FeeHandler::withdraw_fee, self.schedule_fee.asset_location: {:?}, self.schedule_fee.amount: {:?}",
+			self.schedule_fee.asset_location, self.schedule_fee.amount);
+		// Withdraw schedule fee
+		if !self.schedule_fee.amount.is_zero() {
```
chrisli30, line 152: What happens if it is zero? A better structure would be
```
if is_zero() {
log debug or throw error
}
else{
the withdraw logic
}
```

pallets/automation-time/src/fees.rs
```diff
@@ -71,59 +78,110 @@ where
...
+			.and_then(|balance_minus_fee| balance_minus_fee.checked_sub(&min_balance))
 			.ok_or(DispatchError::Token(BelowMinimum))?;
-		T::MultiCurrency::ensure_can_withdraw(currency_id, &self.owner, fee)?;
+
+		T::MultiCurrency::ensure_can_withdraw(currency_id, &self.owner, amount)?;
+
+		Ok(())
+	}
+
+	/// Ensure the fee can be paid.
```
chrisli30, line 106: the separation of the can_pay_fee code logic looks much cleaner 👍

pallets/automation-time/src/fees.rs
```diff
@@ -71,59 +78,110 @@ where
...
+		if let Some(execution_fee) = &self.execution_fee {
+			if execution_fee.is_local {
+				log::debug!(target: "FeeHandler", "FeeHandler::withdraw_fee, self.execution_fee.asset_location: {:?}, self.execution_fee.amount: {:?}",
+					execution_fee.asset_location, execution_fee.amount);
+				let currency_id = T::CurrencyIdConvert::convert(execution_fee.asset_location)
+					.ok_or("InconvertibleMultilocation")?;
 
-				if self.execution_fee_amount > MultiBalanceOf::<T>::zero() {
+				let execution_fee_amount = execution_fee.amount;
+				if !execution_fee_amount.is_zero() {
```
chrisli30, line 174: The same question as above.

pallets/automation-time/src/lib.rs
```diff
@@ -1551,6 +1559,21 @@ pub mod pallet {
 
 			Ok(fee)
 		}
+
+		pub fn ensure_supported_execution_fee_location(
```
chrisli30, line 1563: I like this ensure function, but let’s add function comments to it. Especially, for a function like this you need to mention in the comment that "if the locations can not be verified, an error such as InvalidAssetLocation or UnsupportedFeePayment will be thrown".

chrisli30 (commented):
The code is now much cleaner, making it easier for me to read and understand. 👍

pallets/xcmp-handler/src/lib.rs
```diff
@@ -228,6 +228,32 @@ pub mod pallet {
...
+				Xcm(vec![
+					WithdrawAsset::<<T as pallet::Config>::RuntimeCall>(asset.clone().into()),
+					BurnAsset::<<T as pallet::Config>::RuntimeCall>(asset.into()),
+				])
+			} else {
+				return Err(Error::<T>::UnsupportedFeePayment.into())
+			};
+
+			Ok(local_xcm)
+		}
```
chrisli30, line 255: okay, let’s make the logic simpler with some restructuring. Because the local_xcm and target_xcm always go in pair, let’s not write `if reserve == MultiLocation::here()` twice separately. Instead,

```
let reserve = T::ReserveProvider::reserve(&asset).ok_or(Error::<T>::InvalidAssetLocation)?;
if reserve == MultiLocation::here() {
let local_xcm = ...
let target_xcm = ...
Ok((local_xcm, target_xcm))
} else if reserve == destination {
let local_xcm = ...
let target_xcm = ...
Ok((local_xcm, target_xcm))
} else {}

```

chrisli30 (changes requested):

pallets/automation-time/src/fees.rs
imstar15: When the schedule_fee is not equal to zero, execute the withdrawal process; otherwise, skip it.

pallets/automation-time/src/fees.rs
imstar15: When the execution_fee_amount is not equal to zero, execute the withdrawal process; otherwise, skip it.

pallets/automation-time/src/fees.rs
chrisli30: Please add your explanation to this section of code as a comment.

```
When the expected deduction amount, schedule_fee_amount, is not equal to zero, execute the withdrawal process; otherwise, there’s no need to deduct.
```

pallets/automation-time/src/fees.rs
chrisli30: Please add your explanation to this section of code as a comment.

```
When the expected deduction amount, execution_fee_amount, is not equal to zero, execute the withdrawal process; otherwise, there’s no need to deduct.
```

pallets/automation-time/src/fees.rs
```diff
@@ -267,6 +267,100 @@ mod tests {
 		})
 	}
 
+	#[test]
+	fn call_pay_checked_fees_for_with_normal_flow_and_enough_execution_fee_success() {
+		new_test_ext(0).execute_with(|| {
+			let destination = MultiLocation::new(1, X1(Parachain(PARA_ID)));
+			let alice = AccountId32::new(ALICE);
+			let mut spy = 0;
```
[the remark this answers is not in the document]
chrisli30: let’s call it `has_callback_run` and set it initially false.

pallets/automation-time/src/fees.rs
```diff
@@ -267,6 +267,100 @@ mod tests {
...
+				&action,
+				1,
+				|| Ok(()),
+			);
+			assert_err!(result, Error::<Test>::InsufficientBalance);
+		})
+	}
+
+	#[test]
+	fn call_pay_checked_fees_for_with_alternate_flow_and_no_execution_fee_success() {
```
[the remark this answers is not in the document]
chrisli30: execution_fee in alternate flow is not deducted on our side but after XCM, so there’s no `insufficient_execution_fee` case.

pallets/automation-time/src/fees.rs
```diff
@@ -194,6 +267,100 @@ mod tests {
...
+	#[test]
+	fn call_pay_checked_fees_for_with_normal_flow_and_insufficent_execution_fee_will_fail() {
+		new_test_ext(0).execute_with(|| {
+			let destination = MultiLocation::new(1, X1(Parachain(PARA_ID)));
+			let alice = AccountId32::new(ALICE);
+			fund_account(&alice, 900_000_000, 1, Some(0));
+
+			let action = Action::XCMP {
+				destination,
+				schedule_fee: NATIVE_LOCATION,
```
chrisli30, line 312: Let’s make sure there’s a case covering the scenarion, when schedule_fee is a foreign asset and its balance is insufficient, the code will also throw a InsufficientBalance error.

imstar15: Fixed

pallets/automation-time/src/fees.rs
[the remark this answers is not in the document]
imstar15: Fixed

pallets/automation-time/src/fees.rs
```diff
@@ -19,6 +19,7 @@
 use crate::{AccountOf, Action, ActionOf, Config, Error, MultiBalanceOf, Pallet};
 
 use frame_support::traits::Get;
+use frame_system::RawOrigin;
```
chrisli30, line 22: Why is the addition of this line? Is it necessary?

chrisli30 (approved):
The comments are all addressed.

pallets/automation-time/src/fees.rs
imstar15: Yes.
During testing, we require Root permissions to deposit funds into the test account.

PR #468, opened 2023-11-28
Remove xcmp_handler in benchmark workflow

imstar15:
We have removed the benchmarking.rs file from the xcmp_handler pallet in the following PR because all extrinsic call interfaces (such as set_asset_config) have been removed from this pallet.

https://github.com/OAK-Foundation/OAK-blockchain/pull/366

There is no need to specify running benchmarks for this pallet in benchmark.yml anymore.

Because the benchmarking.rs file no longer exists, an error will occur when GitHub Actions run.

https://github.com/OAK-Foundation/OAK-blockchain/actions/runs/7014798037/job/19083688075

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       pallet:
         description: provided a list of pallet to run benchmark on
-        default: 'automation_time,automation_price,vesting,valve,xcmp_handler'
+        default: 'automation_time,automation_price,vesting,valve'
         required: true
 
 concurrency:


chrisli30:
@imstar15 can you remember to add explanation to your PRs? Why do we need to remove xcm_handler?

imstar15:
> @imstar15 can you remember to add explanation to your PRs? Why do we need to remove xcm_handler?

Sorry. I added explanation.

chrisli30 (approved):
Okay, the explanation is clear. 👌

PR #472, opened 2023-12-01
remove bdnodes reference

v9n:
we already removed on our ansible, but also removed here because this node is dead now and no longer respond anyway.


chrisli30:
The bdnode has already been shutted down, so let’s merge this PR.

Issue #475, opened 2023-12-04
GLIBC not found for Turing node on v2.0.2


jakobilobi:
**Question**

Hi there,
I'm from Dwellir, a node provider running a node on the Turing chain, and when I just now tried upgrading our node from `v2.0.1` to `v2.0.2` I got this error when starting the node client (`oak-collator` client is renamed `polkadot` in our infra) after the upgrade:

```
/home/polkadot/polkadot: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /home/polkadot/polkadot)
/home/polkadot/polkadot: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /home/polkadot/polkadot)
/home/polkadot/polkadot: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /home/polkadot/polkadot)
/home/polkadot/polkadot: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /home/polkadot/polkadot)
/home/polkadot/polkadot: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /home/polkadot/polkadot)
```

I couldn't find any info in the [latest release notes](https://github.com/OAK-Foundation/OAK-blockchain/releases/tag/v2.0.2) about any new requirements for this version. My temporary solution was to downgrade to `v2.0.1` again, which lets the node run fine, but I do have questions:

- Is this new requirement an intended change?
- What should we install to make the `v2.0.2` client work?

Cheers,
Jakob

chrisli30:
Hey Jakob,

Thank you for the reporting. Hmm let us take a look, but the 2.0.2 is a minor upgrade and shouldn’t include any node changes or upgrade requirement. We are testing 2.1.0 node upgrade in staging which is major and require another upgrade next week.

jakobilobi:
Hi there,

Tested with the new binary, works great :+1: 

Minor thing but noticed the client still presents itself as v2.0.0, even though it's the v2.0.2 one, thought you'd might want to know!

```
./oak-collator --version
2.0.0-a2a9cc182f3
```

chrisli30:
[...]

PR #476, opened 2023-12-04
Inspect node compiler

v9n:
Inspect our glibc and log the glibc the binary used to prevent the case where our binary get update due to OS update and we shorted sighted.

ref #475


chrisli30 (approved):
Okay nice, this objdump command will display the version that this binary is using right?

It would be great to add a comment,

```
# For example, expected output is 2.30
```

chrisli30 (commented):
Yea this looks good!

PR #487, opened 2023-12-12
allow xtokens pallet in ScheduleAllowList

v9n:
We're attempting to schedule bridge trasfer, which run through Xtokens so we add it to his list.


chrisli30 (approved):
ah, right ... this makes sense.

PR #488, opened 2023-12-12
bump minor imp_version runtime release

v9n:
This wont' change any existing but only allow add XTokens pallet to our ScheduleAllowList. This is a runtime only upgrade


chrisli30 (approved):
Looks good 🆗

PR #490, opened 2023-12-14
add Currencies and Utility pallet to AllowList

v9n:
in dynamic dispatch, we may use Utility.batch to bundle multiple call into the same task, therefore we would need to allow this.

similarly with currencies, we don't use it yet, but may need down the line and save us another update.


chrisli30 (approved):
LGTM

PR #501, opened 2024-01-09
Polkadot v1.1.0

decentration:
Attempted upgrade to polkadot-v1.1.0 because open web3 stack does not go higher than v1.1.0
https://github.com/open-web3-stack/open-runtime-module-library/tags 

replaced branches for tags

replaced moonbeam and moonkit forks to call directly from moonbeam, for ease. 

Some errors in the custom pallets, xcmp-helper, valve, and primitives. 

Noticed primitives is from Astar mock file, but that is far is have i got so far.


chrisli30:
Thank you for your initiative, Ramsey!

To provide more context regarding this PR: Following the release of the Polkadot SDK version 1.0 and subsequent versions, parachains have been updating their codebases to align with the new Polkadot standards. For our upcoming upgrade of the Polkadot core code, it's essential that we also adopt this new format.

PR #499, opened 2024-01-05
Modify the scheduleAs parameter to be an optional parameter

imstar15:
We can use the Derivative account as the task executor without the need to add proxy relationships on the Turing chain.

PR #503, opened 2024-01-09
Bump to spec version 297

imstar15:
**PRs:**
Modify the scheduleAs parameter to be an optional parameter
https://github.com/OAK-Foundation/OAK-blockchain/pull/499

### DIFF
--- a/runtime/neumann/src/lib.rs
+++ b/runtime/neumann/src/lib.rs
@@ -173,10 +173,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("neumann"),
 	impl_name: create_runtime_str!("neumann"),
 	authoring_version: 1,
-	spec_version: 296,
-	impl_version: 1,
+	spec_version: 297,
+	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 19,
+	transaction_version: 20,
 	state_version: 0,
 };
 
--- a/runtime/oak/src/lib.rs
+++ b/runtime/oak/src/lib.rs
@@ -179,10 +179,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("oak"),
 	impl_name: create_runtime_str!("oak"),
 	authoring_version: 1,
-	spec_version: 296,
-	impl_version: 3,
+	spec_version: 297,
+	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 19,
+	transaction_version: 20,
 	state_version: 0,
 };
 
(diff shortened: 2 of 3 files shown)


chrisli30 (changes requested):
No, why do we change the spec_version of the chain? Please read the instructions on Notion, specifically "**transaction number must change** when an existing dispatchable (module ID, dispatch ID) is changed, either through an alteration in its user-level semantics, **a parameter added/removed/changed**" and "If this number changes, **then `spec_version` must change**, also."

Full reference:
```
	/// Version of the implementation of the specification. Nodes are free to ignore this; it
	/// serves only as an indication that the code is different; as long as the other two versions
	/// are the same then while the actual code may be different, it is nonetheless required to
	/// do the same thing.
	/// Non-consensus-breaking optimizations are about the only changes that could be made which
	/// would result in only the `impl_version` changing.
	pub impl_version: u32,

	/// All existing dispatches are fully compatible when this number doesn't change. If this
	/// number changes, then `spec_version` must change, also.
	///
	/// This number must change when an existing dispatchable (module ID, dispatch ID) is changed,
	/// either through an alteration in its user-level semantics, a parameter
	/// added/removed/changed, a dispatchable being removed, a module being removed, or a
	/// dispatchable/module changing its index.
	///
	/// It need *not* change when a new module is added or when a dispatchable is added.
	pub transaction_version: u32,
```

imstar15:
> No, why do we change the spec_version of the chain? Please read the instructions on Notion, specifically "**transaction number must change** when an existing dispatchable (module ID, dispatch ID) is changed, either through an alteration in its user-level semantics, **a parameter added/removed/changed**" and "If this number changes, **then `spec_version` must change**, also."
> 
> Full reference:
> 
> ```
> 	/// Version of the implementation of the specification. Nodes are free to ignore this; it
> 	/// serves only as an indication that the code is different; as long as the other two versions
> 	/// are the same then while the actual code may be different, it is nonetheless required to
> 	/// do the same thing.
> 	/// Non-consensus-breaking optimizations are about the only changes that could be made which
> 	/// would result in only the `impl_version` changing.
> 	pub impl_version: u32,
> 
> 	/// All existing dispatches are fully compatible when this number doesn't change. If this
> 	/// number changes, then `spec_version` must change, also.
> 	///
> 	/// This number must change when an existing dispatchable (module ID, dispatch ID) is changed,
> 	/// either through an alteration in its user-level semantics, a parameter
> 	/// added/removed/changed, a dispatchable being removed, a module being removed, or a
> 	/// dispatchable/module changing its index.
> 	///
> 	/// It need *not* change when a new module is added or when a dispatchable is added.
> 	pub transaction_version: u32,
> ```

1. The scheduleXcmpTaskThoughProxy dispatch has been removed.
2. The dispatch ID for scheduleXcmpTask has been changed.
3. scheduleXcmpTask has added scheduleAs and instructionSequence parameters.

Hence, the transaction_version and spec_version need to be changed.

chrisli30 (approved):
Looks good to me. The spec_version needs to change because of the above reasons :+1:

PR #506, opened 2024-02-16
Add DescribeBodyTerminal to DescribeAllTerminal

arrudagates:
This PR adds DescribeBodyTerminal to DescribeAllTerminal, keeping the local implementation of DescribeAllTerminal (in the local primitives crate) in sync with later versions of the Polkadot SDK, where DescribeBodyTerminal was introduced in-between versions 1.1.0 and 1.2.0.

For reference, [this](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.2.0/polkadot/xcm/xcm-builder/src/location_conversion.rs#L104) is the implementation in v1.2.0.


chrisli30:
Thank you, Gab! Let us take a look at the changes!

chrisli30:
More context from Gab "The DescribeBody terminal is necessary for the HashedDescription to describe XCM locations with the Plurality junction, which we use in our Saturn accounts"

Lalsha40:
Where are pr no

chrisli30 (approved):
Looks good, and we are ready to run an upgrade.

PR #530, opened 2024-08-16
reword `runtime-dep.py` message

DemoYeti:


chrisli30 (approved):
@DemoYeti looks good and thank you

